Issue #174, opened 2026-05-27
Strudel-official viz methods don't work out of the gate (`._scope()`, `.pianoroll()`, `._punchcard()`, `._tscope()`)


MrityunjayBhardwaj:
## Problem

Pasted standard Strudel code that uses official inline viz methods does **not** work out of the gate. Stave deliberately doesn't load `@strudel/draw` (it injects a fullscreen canvas) and instead re-implements viz inline via `.viz("name")` + a capture-wrapper loop in `StrudelEngine.ts:532-580`. Measured against Strudel's actual vocabulary (`@strudel/draw` + `@strudel/webaudio` source), four breaks:

| Form | Current | Cause |
|------|---------|-------|
| `._tscope()` | throws | `tscope` missing from list |
| `.pianoroll()` / `.punchcard()` / `.spiral()` / `.pitchwheel()` / `.wordfall()` | throws | non-underscore forms never defined (draw not loaded) |
| `.scope()` / `.tscope()` / `.fscope()` / `.spectrum()` | draws a stray FULLSCREEN canvas | `@strudel/webaudio` index side-imports `scope.mjs`/`spectrum.mjs` → real `.analyze().draw()` → the canvas Stave avoided |
| `._punchcard()` / `._tscope()` | silent no-viz | captured but `resolveDescriptor` returns undefined (no renderer) |

## Official Strudel viz vocabulary (verified against source)

`pianoroll, punchcard, spiral, pitchwheel` (from `@strudel/draw`), `scope, tscope (=scope alias), fscope, spectrum` (from `@strudel/webaudio`), `wordfall` (draw). `markcss` is a CSS highlight helper, not a canvas viz.

## Fix (single boundary: the viz-capture wrapper in `StrudelEngine.ts`)

1. Canonical vocab → renderer map (single source of truth) with aliases `tscope→scope`, `punchcard→pianoroll` (real PunchcardSketch = follow-up).
2. Intercept BOTH `name()` and `_name()`; each sets `_pendingViz` + returns `this`. Do NOT chain to Strudel's real method (prevents the fullscreen canvas on `.scope()` etc.).
3. Restore descriptors symmetrically post-eval.
4. Drop `markCSS` from the viz list.

Safe because `.viz("scope")` already renders today without Strudel's `.analyze()` (Stave taps its own per-orbit analysers).

## Observation gate
[...]

PR #178, opened 2026-05-27
feat(viz): Strudel-official methods + Piano Roll single source of truth (closes #181, #182, #183)

### LINKED ISSUE #181: Named-viz registry collision: scope.p5 and scope.hydra both register as "scope" → inline .viz("scope") is nondeterministic (p5 vs hydra)
Found while testing #174. `flushToPreset` sets a preset's name to the file basename (vizPresetBridge.ts:198), and `registerPresetAsNamedViz` keys the named-viz registry by `preset.name`. Both `preset/viz/scope.p5` and `preset/viz/scope.hydra` → name **"scope"**, so `registerAllVizFiles` (StrudelEditorClient) registers them under the same key — **last-write-wins** in arbitrary `listWorkspaceFiles()` order.

Result: inline `.viz("scope")` / `._scope()` resolves via the named registry (checked first in resolveDescriptor) to whichever scope preset registered last — possibly the **hydra** one — while the #174 backdrop `.scope()` deterministically picks `scope.p5`. They render different viz.

The `mode:renderer` convention says the bare mode name (`scope`) belongs to the default renderer (p5), with `scope:hydra` for the variant.

**Fix:** in `registerAllVizFiles`, when a p5 file and a hydra file share a basename, register the hydra one under `<name>:hydra` so the bare name stays the p5 default. Surgical — only de-collides actual clashes; hydra-only viz (kaleidoscope) keep their bare name.

### LINKED ISSUE #182: Inline .viz("pianoroll") renders built-in PianorollSketch, .pianoroll() backdrop renders Piano Roll.p5 preset — visually different
Sibling of #181, different divergence point. The bundled file is **`Piano Roll.p5`** (basename "Piano Roll"), but inline lookups use the descriptor id **"pianoroll"** (lowercase). `registerAllVizFiles` keyed the named registry by the display basename ("Piano Roll"), so inline `.viz("pianoroll")` MISSES the named registry and falls to the built-in `PianorollSketch`. The #174 `.pianoroll()` backdrop normalizes "Piano Roll"→"pianoroll" and renders the `Piano Roll.p5` file. Two different implementations → visibly different blocks/positions/active-highlight.

(For scope it lined up only because `scope.p5`'s basename "scope" already equals the descriptor id.)

**Fix:** register each viz preset under BOTH its display basename AND its normalized form (lowercase, alphanumerics), so inline `.viz("pianoroll")` and `.viz("Piano Roll")` both resolve to the `Piano Roll.p5` preset — the same source the backdrop uses. Generalizes the #181 hydra-suffix rule.

### LINKED ISSUE #183: Viz single source of truth (Model B): inline .viz("pianoroll") must render the same preset as the .pianoroll() backdrop
## Problem
Every standard Strudel viz exists twice: a built-in TS sketch (`DEFAULT_VIZ_DESCRIPTORS` → `PianorollSketch`/`ScopeSketch`/…) and an editable preset file (`Piano Roll.p5`, `scope.p5`, …). The named-viz registry already collapses this for 6/7 viz — `scope`, `fscope`, `spectrum`, `spiral`, `pitchwheel`, `wordfall` all resolve inline to their preset because each preset name equals its canonical id.

**Piano Roll alone diverges**: its preset is named `"Piano Roll"` but inline resolves the id `"pianoroll"` → named-registry miss → falls through to the built-in `PianorollSketch` TS class; the backdrop normalizes `"Piano Roll"→"pianoroll"` and renders the file. Result: inline ≠ backdrop for pianoroll. Catalogue P73 / PV56.

## Fix (Model B — editable preset files are the single source of truth)
1. **Make inline resolve Piano Roll to its preset** — normalize the named-registry lookup so `.viz("pianoroll")` matches the registered `"Piano Roll"` preset (the same normalization the backdrop already uses: `toLowerCase().replace(/[\s\-_]/g,"")`). NOT the per-pointer register-under-id patch that regressed earlier (empty buffered scheduler at mount + churn).
2. **Retire the built-in TS sketch classes** so presets are the only implementation (satisfies PV56).
3. Verify by screenshot: inline == backdrop for pianoroll, scope, spectrum, spiral, pitchwheel, wordfall.

Folds into #178.

MrityunjayBhardwaj:
Folds the original `#174` Strudel-official viz methods (`._name` / `.name`) together with the Model B single-source-of-truth work that finishes the inline-vs-backdrop story.

## What's in this PR (8 commits)

**Strudel viz methods (#174):**
- `._name()` (underscore) → inline viz zone (Strudel's own semantic).
- `.name()` (non-underscore) → Stave backdrop (pinned, "set bg" indicator).
- Removing the non-underscore method clears the backdrop — code is the source of truth.
- Never injects Strudel's own fullscreen `#test-canvas`.

**Scope single source (closes #181):**
- `scope.p5` + `scope.hydra` were last-write-wins under the same name. Hydra now registers as `scope:hydra` when it shares a basename with a p5 file, so the bare mode reserves the p5 default and `.scope()` inline + `.scope()` backdrop render the identical p5 line-scope.

**Piano Roll single source — Model B (closes #182, #183):**
- The viz duality (built-in TS sketch vs editable preset file) was already collapsed for 6/7 standard viz via the named-viz registry. Piano Roll alone diverged because its preset is named `"Piano Roll"` while inline resolves the id `"pianoroll"` → registry miss → fell through to the built-in.
- `namedVizRegistry` gains a normalized-name index (lower + strip `\s\-_`, mirroring the backdrop's normalization). `resolveDescriptor` tries exact `getNamedViz` first (preserves shadowing), then `getNamedVizByNormalized`. Inline `.viz("pianoroll")` now reaches the `"Piano Roll"` preset — the same source the `.pianoroll()` backdrop renders.
- The bundled `PIANOROLL_P5_CODE` was lower-fidelity than the built-in it replaces (raw note/127 roll, no fold lanes, NaN on note names like `"c"`). Rewrote it as a faithful port of the built-in's rich structure with the bonus of wiring up the latent drum/bass/pad classification — fold-by-pitch lanes, classified colors (drums=orange, bass=cyan, pad=green, melody=purple), active-note brightening + outline, gain×velocity alpha, playhead line, robust note-NAME → MIDI.

**Catalogue:**
- **P73** resolved (with framing correction — the prior "register-under-id blanked BOTH" was misattributed; observation revealed the real cause was the preset's NaN bug).
- **P74** added — a genuinely new pattern: "consolidation downgrades silently when the surviving implementation is lower-fidelity than the one it replaces; observation is a HARD gate on consolidation."
- **PV56** status: satisfied at resolution + render. Physical retire of `DEFAULT_VIZ_DESCRIPTORS` p5 entries tracked as **#184**.

## Test plan

- Editor unit: 1648 passed (+8 normalized-lookup tests).
- App unit: 430 passed.
- e2e: 6 passed — 4 existing `strudel-viz-methods.spec.ts` + 2 new `viz-single-source.spec.ts` (non-blank assertion for inline `.viz("pianoroll")` + `.pianoroll()` backdrop using note-name patterns, which used to NaN-blank under the simpler preset).
- Real `next build` green (P72 — `pnpm --filter @stave/app build` exit 0; `BUILD_ID` present).
- **Direct observation** (Lokāyata): screenshots captured for the new spec confirm inline and backdrop render the identical rich roll on a melodic pattern AND a `stack(note, drums)` pattern (orange drum lanes + purple melody, active-note highlight, playhead line). Zero NaN warnings from the preset's draw loop.

## Follow-ups

- **#184** — Retire the built-in p5 TS sketch classes (`DEFAULT_VIZ_DESCRIPTORS` should be the bundled preset code compiled in-package). Currently dead in the app; physical removal closes PV56 everywhere.
- **#179** — moot under Model B (`PunchcardSketch` follow-up).
- **#180** — Ctrl+Enter while playing is a silent no-op; re-eval needs stop+play.

Closes #181, closes #182, closes #183.


MrityunjayBhardwaj:
**Refinement landed** ([2586598](https://github.com/MrityunjayBhardwaj/stave-code/pull/178/commits/2586598)) per the discuss decision: **code is the source of truth.** Removing a non-underscore viz method from code now clears the backdrop on the next eval. Manual picker selection is overridden when code has no viz method — by design; the backdrop state effectively lives in the (persisted) code.

**All gates green:**
- 4 e2e cases in `strudel-viz-methods.spec.ts` pass (pin, normalized-basename, clear-on-removal, underscore-inline)
- editor 1640 + app 430 vitest pass (no regressions)
- `pnpm turbo run build --filter=@stave/app` ✓ (real next build, full tsc)

**Surfaced separately:** [#180](https://github.com/MrityunjayBhardwaj/stave-code/issues/180) — Ctrl+Enter while playing is a silent no-op (re-eval needs stop+play or live mode). Doesn't block this PR (live mode + first-play still work as the user expects).

Issue #199, opened 2026-05-29
File-history Phase I — manual commits + labels (exempt from retention pruning)


MrityunjayBhardwaj:
Phase I of the file-history milestone. User-initiated, named checkpoints on top of the auto-commit stream.

## Goal

Let the user create a manual commit with a label ("before the big refactor", "v1 demo state") at any point. Manual commits — and the seed commit `c0` — are **exempt from retention pruning** (Phase E policy), so they're permanent anchors.

## Scope

- "Commit now" affordance in the History panel (Phase G) — captures current project state via `commit(projectId, {kind:'manual', label})`.
- Label input (inline, validated non-empty); labels shown on the graph nodes.
- Pruning respects `kind:'manual'` and `kind:'seed'` → never pruned (assert in Phase F's retention driver — this issue adds the UI + the test that proves exemption).
- Optional: rename/edit a commit's label after creation.
- Playwright spec (PV60-gated): create manual commit with label → run pruning (simulate time / force prune) → assert manual commit survives while old auto-commits are pruned.

## Dependencies

Blocked by Phase F (#196, provides `kind` + retention driver). Surfaces in Phase G (#197) panel.

_Part of the file-history milestone (plan locked 2026-05-29)._

Issue #195, opened 2026-05-29
File-history Phase E — RESEARCH: auto-commit cadence, retention, significance threshold


MrityunjayBhardwaj:
Phase E of the file-history milestone. Research + decision phase that locks the policy knobs before Phase F builds the commit store. **Deliverable: `.planning/phase-E-history/RESEARCH.md`.**

## Knobs to lock (recommendations from the locked plan)

| Knob | Recommendation to validate |
|---|---|
| **Auto-commit cadence** | Idle-debounced 5s **+** per-save **+** significance threshold (only auto-commit when ≥5 changed lines OR ≥200 changed chars since last commit). Avoids a commit per keystroke and a commit storm on bulk paste. |
| **Retention** | 24h: keep all auto-commits · 30d: keep daily · older: keep monthly · **manual commits + seed (c0) exempt** from pruning. |
| **Commit content form** | Only-changed-files snapshot per commit (NOT Yjs deltas) — simpler, no replay chain, `getFileContentAt` walks back to nearest writer. |

## Research questions

1. IDB write cost at the proposed cadence across a multi-file project — does idle-5s + significance keep writes bounded? Measure against `stave-snapshots` current footprint.
2. Significance heuristic: char/line diff vs structural (AST) — char/line is the cheap default; confirm it doesn't miss meaningful small edits (e.g. a 1-char param change in a `.viz`).
3. Retention pruning algorithm — bucketed (hour/day/month) vs reservoir; interaction with `fileIndex` integrity (pruning a commit must not orphan a file's only copy → keep-if-sole-writer rule).
4. Interaction with existing `MAX_AUTO_SNAPSHOTS=10` in `snapshotStore.ts` — migrate or run alongside?

## Output

`RESEARCH.md` with each knob locked + rationale, the pruning algorithm spec, and the `keep-if-sole-writer` invariant for `fileIndex` (candidate for a new PV entry).

## Dependencies

Blocked by Phase A (#189, shipped #194). Feeds Phase F.

_Part of the file-history milestone (plan locked 2026-05-29). Siblings: #191 (B), #193 (D), and the Phase F/G/H/I issues._

Issue #189, opened 2026-05-28
StrudelEditorClient seedPresets clobbers user edits to bundled Piano Roll preset on every load


MrityunjayBhardwaj:
Surfaced while doing #184. Out of scope for that PR.

`packages/app/src/components/StrudelEditorClient.tsx:278-304` runs a `seedPresets` effect on every mount that unconditionally `VizPresetStore.put`s the bundled Piano Roll preset with `code: PIANOROLL_P5_CODE` and `code: PIANOROLL_HYDRA_CODE` (plus name / renderer / requires), overwriting whatever the user has saved to the persisted `VizPresetStore` entry for the bundled IDs.

```ts
await VizPresetStore.put({
  ...existingP5,
  id: seedState.p5PresetId, name: "Piano Roll", renderer: "p5",
  code: PIANOROLL_P5_CODE, requires: ["streaming"],  // ← clobbers user edits
  ...
});
```

The `{ ...existingP5, ..., code: PIANOROLL_P5_CODE }` shape implies an intent to preserve existing fields but explicitly REPLACE code — which is the opposite of the "editable / forkable preset" promise (the bundled preset should be a starting point the user can modify).

## Repro

1. Open the bundled Piano Roll preset (`preset/viz/Piano Roll.p5`).
2. Edit it (change a color, add a line).
3. Reload the page.
4. Edits gone — restored to `PIANOROLL_P5_CODE`.

## Proposed fix

Either:
- Seed only when no existing entry (`if (!existingP5) await put({...})`) — bundled becomes a one-time scaffold; subsequent loads honor user state.
- Or seed name/renderer/requires/nativeSize metadata but leave `code` alone when an entry exists.

The second is what the spread shape suggests was originally intended — the `code` line is the bug.

## Scope

This is the bundled Piano Roll only. The other 6 p5 viz files (`scope.p5`, `fscope.p5`, `spectrum.p5`, `spiral.p5`, `pitchwheel.p5`, `wordfall.p5`) are seeded by the workspace file template path (`templates.ts:makeFullStarterFiles`) which goes through `seedWorkspaceFile` and respects existing user content. So only Piano Roll has this clobber.

Source: #184 retrospective on the build path.

MrityunjayBhardwaj:
[...]

PR #200, opened 2026-05-29
:memo: docs(catalogue): P78–P79 multi-writer regression traps + PV60 stash-and-verify gate

MrityunjayBhardwaj:
## Summary

Captures the diagnostic knowledge surfaced during the #189/#194 session (seedPresets clobber fix + its regression test). Three catalogue entries that codify regression-test failure modes for multi-writer stores — directly load-bearing for the file-history milestone (#195–#199), where the Yjs/Monaco/named-viz/VizPreset write choreography is the same shape that bit #189.

- **P78** (hetvabhasa) — *Multi-writer regression test trap.* When a store has N writers, a fix to one writer must be probed at the layer/timing that captures their **combined** post-state. Worked example: the three `VizPresetStore` writers in `StrudelEditorClient` (`registerAllVizFiles` flush · `seedPresets` · `useVizRefWatcher`). Documents both failed test attempts (wrong layer, wrong timing) and the real fix path.
- **P79** (hetvabhasa) — *A regression test that doesn't FAIL pre-fix is vestigial.* Green-both-ways proves nothing; it'll be green next time the bug recurs too.
- **PV60** (vyapti) — *Stash-and-verify gate.* Every regression test paired with a fix MUST be shown to fail when the fix is stashed, then pass when restored. The oscillation goes in the commit body. **Applies to every regression test in the file-history milestone.**

## Test plan

Docs-only — `.anvi/` catalogue markdown. No code change, no build impact. REFs trace to PR #194 (#189) and the named source files.

## Notes

`.anvi/` is gitignored locally; committed with `git add -f` per the catalogue-as-living-docs convention (same as #187).

### DIFF
--- a/.anvi/hetvabhasa.md
+++ b/.anvi/hetvabhasa.md
@@ -2767,3 +2767,98 @@ delta is computed against, not from a subset of it.
 
 **REF:** PR #185 (#175) — `StrudelEditorClient.tsx:prevFileIdsRef` seed
 correction; see PV59.
+
+## P78 — Multi-writer regression test trap: fix verified at the wrong layer / wrong timing because N writers exist
+
+**Pattern:** A store is written by N independent code paths. A bug
+exists at writer-A (e.g. clobber-on-every-mount). The fix removes the
+clobber at writer-A. The regression test attempts to probe the fix by
+either (i) probing a layer ABOVE the store (UI text, editor pane) and
+expecting the store-level fix to show through, or (ii) probing the
+store directly but at a TIMING SLICE where writer-B's legitimate sync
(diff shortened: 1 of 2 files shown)


MrityunjayBhardwaj:
Closing — decided to stop tracking `.anvi/` catalogues in the repo rather than PR catalogue updates. The directory is gitignored (.gitignore:11); the three files (hetvabhasa/krama/vyapti) were tracked from before that rule and are being untracked via `git rm --cached`. Catalogues remain on disk as the project's living diagnostic docs (the hook reads `cwd/.anvi` directly); they just leave git history. P78/P79/PV60 content preserved on disk.

PR #229, opened 2026-06-03
⚡ perf: profiler — measure viz/scheduler cost before optimizing (#228)

### LINKED ISSUE #228: perf: build a profiler to ground viz/scheduler optimization (observe before optimize)
## Problem
We want to optimize runtime performance (viz smoothness, main-thread budget, scheduler tick latency) but have **zero instrumentation** today. Optimizing now = inference, not observation. We need real numbers first.

Grounded architecture (so we measure the right things):
- Audio DSP is already off-thread (Web Audio / AudioWorklet).
- Scheduler is main-thread `setInterval` lookahead (`@strudel/core` `zyklus.mjs:38`) — jank-resilient up to the lookahead window.
- Viz is main-thread `requestAnimationFrame` — **multiple instances** (inline per-codeblock + backdrop + popout), each its own loop. This is the real main-thread competition.
- SignalBus does per-frame work on main: `scheduler.query` (combined + per-track) and `getByte*Data` per analyser, every frame.

## Solution
A full-fledged, zero-cost-when-disabled profiler:
- Core `Profiler` (ring-buffer stats: count/mean/p50/p95/p99/max per section; per-instance frame interval → FPS + drops; counters; `PerformanceObserver` longtask capture).
- Instrument the real per-frame sites: p5 `__tick`, hydra `pumpAudio` (bus tick/refresh/readAudio split out; hydra `.tick` draw cost), renderer mount/destroy (live-instance counts), engine `onTrigger` (trigger rate).
- A live overlay (toggle via setting + keyboard), throttled refresh.
- A scriptable `window.__stavePerf` snapshot/reset hook so Playwright can drive it and READ numbers (the OBSERVE step).
- Gated by `__STAVE_PERF__` / a setting — inert in normal use.

## Acceptance
- Run against a heavy patch (multiple codeblocks + backdrop + audio), read p50/p95 frame time, per-section ms, longtask count, live-instance counts. The data decides the next optimization (likely OffscreenCanvas-worker for viz, or a targeted bus hotspot — TBD by observation, not assumed).

MrityunjayBhardwaj:
Closes #228.

## Why
We want to optimize runtime performance, but had **zero instrumentation** — any optimization would be inference, not observation. This builds the measurement first.

Grounded correction to the threading mental model (from the source):
- **Audio DSP is already off the main thread** (Web Audio / AudioWorklet).
- The **scheduler** is a main-thread `setInterval` *lookahead* clock (`@strudel/core` `zyklus.mjs:38`) — it schedules audio nodes ~ahead with absolute `AudioContext` times, so it's **jank-resilient** up to the lookahead window. It can't trivially move to a worker because `onTrigger` creates Web Audio nodes (needs the main-thread `AudioContext`).
- The real main-thread competition is the **viz** (`requestAnimationFrame`) — multiple instances (inline + backdrop + popout), each its own loop — plus the SignalBus per-frame work.

## What
A full-fledged, **zero-cost-when-disabled** profiler:
- **Core** (`perf`): ring-buffer section stats (count/mean/p50/p95/p99/max), per-instance frame interval → fps + drops, cumulative **counters** vs live **gauges**, `PerformanceObserver` longtask capture. Every hot-path method early-returns on one boolean when off.
- **Instrumentation**: p5 `__tick` + hydra `pumpAudio` (bus work + hydra draw timed), renderer mount/destroy (live gauges + frame cleanup), engine `onTrigger` (trigger rate). The SignalBus stays **pure** (P12/PV65) — timing wraps the renderer's calls.
- **Overlay HUD** (`PerfOverlay`): polls `snapshot()` at 5Hz only while enabled; shows fps/p95/drops, section p50·p95, longtasks, live counts, trigger rate. Renders null when off.
- **Toggle**: Settings row + **Alt+P** + persisted `stave:perfEnabled` + `window.__stavePerf` / `__STAVE_PERF__` for automation.
- **Observation e2e**: enables the profiler, drives a heavy 4-track + backdrop-viz patch, reads `snapshot()`, prints it, asserts structure.

## First observation (real run — 4-track + pianoroll backdrop, audioCtx running)
```
p5.bus  mean 0.24ms / p95 0.40ms   (~2.4% of a 16.7ms frame — negligible)
p5#6    59.9 fps / p95 18ms / 0 drops
longtasks 0 · triggers 15/s · viz.p5 gauge 1
```
→ For this patch the SignalBus is **not** a bottleneck and the frame is healthy. The OBSERVE step also caught a real measurement bug (gauges must survive `reset()`), now fixed. The open question the data points at next: **multi-instance / hydra-WebGL** scenarios — that's the next round of observation before any optimization (likely OffscreenCanvas-worker for viz, but not committed to until measured).

## Test plan
- Editor unit **1844** green (profiler 14: percentiles, ring eviction, drop detection, disabled no-op, gauge-vs-counter, reset preserves gauges).
- App unit **448** green.
- App e2e: `perf-profiler.spec.ts` green (real snapshot under load).
- Editor + app builds clean.

## Notes
- Profiler is **inert by default** — gated; no overhead in normal use.
- This PR only *measures*. No optimization yet — that's deliberate (observe first).


MrityunjayBhardwaj:
## Critical self-review

**Lint:** fixed 2 new `set-state-in-effect` errors I introduced in `PerfOverlay` (now uses `useSyncExternalStore` + timer-callback setState). `EditorSettingsModal:184` is a pre-existing violation in the settings-seed effect — my one added line follows the same established pattern, so no new error count; out of scope to refactor here.

**Correctness of the measured data:** sound. Sections/frames/counters/gauges are accurate; the OBSERVE step already caught + fixed the one real bug (gauges must survive `reset()`).

**Gaps (all LOW/INFO — filed as #230):**
- p5 user-draw isn't a section (only `p5.bus` + fps); hydra times `hydra.draw`. Asymmetric.
- Sections aggregate across instances (frames are per-instance, so attribution still works).
- Gauge can read negative under StrictMode (construct→destroy without mount); cosmetic.
- Gauge stale if profiler enabled *after* a viz mounted — enable at boot for accurate gauges.
- `p5.bus` begin/end isn't try/finally (bus is pure, won't throw; hydra.draw already is).

**Scope honesty:** this PR only *measures* — no optimization. That's deliberate. The first observation (4-track + backdrop p5: bus p95 0.40ms, 60fps, 0 longtasks) says the bus is negligible and that single-instance frames are healthy; the next step is to observe **multi-instance + hydra-WebGL** scenarios before choosing an optimization (OffscreenCanvas-worker for viz is the leading candidate but unconfirmed).

Issue #488, opened 2026-06-20
Song timeline: a SECOND identical $: arrange track explodes the row count (period correct, rows wrong)


MrityunjayBhardwaj:
## Symptom (observed, not yet diagnosed)

A single arrange track renders correctly, but adding a SECOND identical `$:` arrange track explodes the Song-view row count.

**Repro:**
```strudel
$: arrange([1, s("bd*4")], [2, s("cp sn")], [2, s("hh*8")])
```
→ Song timeline shows **5 cycles** (correct: period = 1+2+2), reasonable rows.

Add an identical line below it:
```strudel
$: arrange([1, s("bd*4")], [2, s("cp sn")], [2, s("hh*8")])
$: arrange([1, s("bd*4")], [2, s("cp sn")], [2, s("hh*8")])
```
→ Song timeline shows **~20 rows** (row explosion). The period/cycle width still looks right; it's the lane/voice ROW count that blows up.

## Notes / where to look (next session — OBSERVE first, don't infer)
- Two identical `$:` tracks share the same sample names (`bd`/`cp`/`sn`/`hh`) but should have distinct `trackId` per `$:` ordinal. Suspect a **laneKey / voice-partition collision or cross-product** between the two tracks (the #424 per-voice sub-row layout, or `laneKeyOf` = `trackId ?? s ?? '$default'`).
- 20 rows is an odd count — likely a multiplicative explosion (per-arm × per-voice × per-track), not a simple double. Dump the actual `scene.lanes` + each lane's `voices`/sub-rows in a live browser probe before forming a hypothesis (P193/P195).
- **Likely PRE-EXISTING** (not from #487 grow+autoscroll — that only touched the extend gesture + marks-over-loopCycles, identical at rest). Verify against the `studio_v0.2.0` baseline early to confirm it's not a regression.
- Related: #485 (residual `$:` ordinal slot churn), #480 (voice sub-row stable order), #424 (per-voice sub-rows).

Deferred from the manual-testing session that reworked #487.

MrityunjayBhardwaj:
Resolved by #492, merged into `studio_v0.2.0`. (Auto-close didn't fire because the PR merged into the studio branch, not the default branch.)

Issue #487, opened 2026-06-20
Song timeline: can't extend the LAST clip — no trailing timeline to drag into


MrityunjayBhardwaj:
## Problem (observed live)

In the Song view you can trim/extend a clip by dragging its right edge, but the **last** clip can't be EXTENDED: its right edge sits at the very end of the timeline, with no empty space to drag into.

## Root cause

The content space is exactly `displayCycles = displaySpan(analysis) = max(1, period ?? horizon)` cycles wide (`FullSongTimeline.tsx:150`). The last clip's `endCycle === displayCycles`, i.e. its right edge is at the right wall of `contentWidth`. The trim drag reads `cyc = xToSongCycle(contentX, displayCycles, cw)`, which caps at `displayCycles`, and at zoom=1 there's no scroll headroom — so `newEnd` can never exceed the current span. There is literally no timeline past the song to extend into.

## Fix

Render trailing **headroom** (a few empty cycles after the song) so the last clip's edge isn't at the wall — the DAW-standard "empty bars after your arrangement" affordance. Headroom changes ONLY the visual span:

- `loopCycles` (the true period) still drives the playhead loop-wrap, note-mark collection, and the bare-track implicit clip's end.
- `displayCycles = loopCycles + headroom` drives the shared cycle→x transform (component AND canvas, in lockstep — PV116), the ruler, and the scroll content width.

After extending, re-eval grows the period → more headroom appears, so extension is unbounded across edits.

MrityunjayBhardwaj:
Resolved by #496, merged into `studio_v0.2.0`. (Auto-close didn't fire because the PR merged into the studio branch, not the default branch.)

PR #493, opened 2026-06-21
✨ Bare loop: selectable, splittable clip + gap-delete (#489, #491)

### LINKED ISSUE #489: Song timeline: select & arrange bars of a bare loop (DAW model)
## Problem

A bare track loop (e.g. `\$: s(\"bd*4\")`) can't be arranged by mouse in the Song view. It's a 1-bar loop (period 1) shown as a single bar, and that bar is a bare *implicit* clip with no `arrange` arms — so it isn't selectable and every clip op (Split/Delete/Duplicate) no-ops (they all require `detectArrangeAt` to find a combinator). Removing the old drag-to-wrap gesture (#488) left bare loops with **no** mouse path into an arrangement.

## Goal (DAW model)

Treat a bare loop like a looped region in a DAW arrangement: show it **tiled across multiple selectable bars**, and let an explicit op on a selected bar **materialize the arrangement**:

- A bare track renders as N selectable bars (the loop tiled). Default to a sensible minimum (so there's room to arrange) and grow via the existing extend gesture (#487).
- **Select** a bar → highlight (like an `arrange` arm).
- **Delete** a selected bar → rewrite `pattern` → `arrange([…, pat], [1, silence], […, pat])`, carving a gap at that bar (the explicit, deliberate \"introduce the combinator\" — reuses the kept `wrapBare`/arrange serializers).
- **Split / Duplicate** a bar → materialize + apply.

Move stays an identity for a uniform loop (#488) — this is the *explicit* arrangement entry-point that replaces the removed accidental drag.

## Notes
- Grounded in DAW arrangement behavior (Ableton/Logic looped-region editing).
- Touches the shared cycle→x transform basis (P197) and the span invariant (PV133 — bare loop display length vs. period).

### LINKED ISSUE #491: Song timeline: deleting a clip should leave a GAP (silence), not ripple-shorten
## Problem
Deleting a clip in the Song timeline currently calls `removeArm` — it removes the arm and pulls later clips left, SHORTENING the arrangement. Grounded in haps: `arrange([1, s("bd*4")], [2, s("bd*4")])` (a delete-middle result) plays `[4,4,4,4]` — no gap.

This is the *ripple* behavior. DAW-standard plain **Delete** leaves a GAP — the arrangement timeline is absolute, so a deleted clip becomes empty space and later clips stay put (Ableton Live, Logic Pro). Ripple/close-gap is a separate, less-common command (Live "Delete Time", etc.).

Surfaced while finishing #489: a user splitting a bare loop and deleting a piece expects a silent bar, but gets a shorter loop.

## Fix
Delete replaces the arm's pattern with `silence`, keeping its cycle width: `[n, pat] → [n, silence]`. Grounded: `arrange([2, s("bd*4")], [2, silence])` plays `[4,4,0,0]` — gap kept in place. New `silenceArm` serializer (pure string surgery, PV123) wired into `handleDeleteClip`; `removeArm` kept as library surface for a future ripple-delete op.

Scope: ALL arrange/cat clips (one coherent model), not just bare-materialized loops.

## Acceptance
- Selecting a clip + Delete leaves silence in its place; later clips do NOT shift; total arrangement length unchanged.
- haps confirm the gap (silent cycles where the clip was).
- Deleting an already-silent clip is a no-op; silencing every arm (empty/muted track) is allowed (valid `arrange([n, silence])`).

MrityunjayBhardwaj:
Closes #489. Closes #491. Stacked on #492 (the #488 no-wrap fix) — **merge #492 first**, then this retargets to `studio_v0.2.0`.

A DAW-model arrange story for a bare loop, in two parts.

## #489 — a bare loop is a selectable, splittable clip
After #488 removed drag-to-wrap, a bare loop (`$: s("bd*4")`) had no mouse path into an arrangement. Reframed (per the user) as: a bare loop is ONE clip spanning the song that you SPLIT to get individually-addressable pieces — DAW-natural, not N pre-cut bars.

- **Display span (D3):** a pure bare loop floors to `MIN_BARE_SPAN` (4) so its one clip has room to split; a real arrangement keeps its own length. "Bare" is probed from the collected events' `armIndex` (via `collectCycles` — correct in both app and mocked tests), not the raw IR shape. The authoritative span flows into `buildTimelineScene` so the clip's `endCycle` and the geometry transform stay in lock-step (revises PV133 for bare tracks).
- **Selectable, not movable:** `clipBodyAt` re-includes the bare clip for SELECT; the move path still excludes it (#488). Only **S** (split) acts on a bare clip — Delete/Duplicate of the whole loop are out of scope (split first).
- **Split materializes:** `materializeBareSplit` → `pat` becomes `arrange([k, pat], [span−k, pat])` (identical sound, two arms). Haps-grounded (`arrange([2,pat],[2,pat])` → `[4,4,4,4]`).

## #491 — Delete leaves a gap (silence), not a ripple
Deleting a clip used `removeArm` (drops the arm, pulls later clips left → shortens). DAW-standard plain Delete leaves a GAP (the timeline is absolute). New `silenceArm` (`[n, pat] → [n, silence]`, width kept) wired into `handleDeleteClip`; `removeArm` kept for a future ripple-delete. Grounded: delete arm 0 of `arrange([2, s("bd")], [2, s("hh")])` → `arrange([2, silence], [2, s("hh")])` → `[4,4,0,0]`.

So the full story: `s("bd*4")` → **S** → `arrange([2, s("bd*4")], [2, s("bd*4")])` → select a piece + **Delete** → `arrange([2, s("bd*4")], [2, silence])` (a real silent bar).

## Test plan
- Editor: `materializeBareSplit` + `silenceArm` unit tests; haps grounding (split is identical; delete is a gap). Editor suite 2394 green; dist rebuilt.
- App: new `full-song-bare-arrange.spec` (select → S → materialize, discriminated); `full-song-arrange-delete` / `clip-ops-e2e` / `clip-chain` delete specs flipped to in-place silence; 109 jsdom timeline tests green; no new tsc errors.
- All position-fraction-sensitive specs re-run (P197); 2 load-flakes confirmed passing in isolation.


MrityunjayBhardwaj:
Merged into `studio_v0.2.0` via merge-commit **e50f002** (composing #489 select/split + #491 gap-delete with #481 voice-order + #487 extend on the integration branch). All commits verified as ancestors of studio by `git merge-base`. The base-retarget couldn't apply because the commits were already in studio. Issues #489/#491 stay open until studio→main.

Issue #527, opened 2026-06-23
Remove the Mixer inspector section + revert click-to-select → click-toggle (redundant in a step grid)


MrityunjayBhardwaj:
**Why:** the selected-note inspector (#432) and the click-to-select gesture it required are **Logic-by-imitation**. In a step-quantized grid the inspector duplicates what's already there:
- **pitch** = the row, **position** = the column, **velocity** = the bar's height + colour (#428) — all already *visible* on the grid; the inspector just re-prints them as text.
- **editing** is already done by gestures: click = add/delete, drag = move, edge-drag = resize, vertical-drag = velocity.

Click-to-select was introduced *only* to feed the inspector, and it overloaded the natural click-toggle (add/remove) gesture.

**What to do (surgical):**
1. Remove the inspector section from the Mixer (`InspectorPanel.tsx` usage in `Mixer.tsx`). **Keep** the Mixer's param knobs, sound/kit pickers (#514/#515), snap dropdown (#432 Slice 2), and quick-transforms — those edit chain args (`.gain`/`.lpf`/`.sound`/`.bank`) that have **no** grid gesture and are *not* redundant.
2. Revert plain click back to **toggle** (add/delete): Piano Roll click-empty = place, click-note = delete; Sequencer click = toggle on/off. Restores the pre-#432 direct-edit model.
3. Drop the `selected` plumbing from `PatternPanel` + the Delete-key removal path that replaced toggle. Trim `inspector.ts` helpers that nothing else uses (keep `gain↔velocity` if copy/paste reuses it).
4. Keep **velocity→color** — it's the visible velocity readout the inspector was duplicating.

**Coordinate with** the copy/paste work: copy/paste keeps its own additive **shift-click / marquee** selection (independent of plain-click), so removing the inspector's single-select doesn't break it.

**Gate:** all grid drag/place/delete/velocity e2e stay green; update the #432 e2e that assert the inspector.

PR #537, opened 2026-06-24
✨ ×2 / ÷2 grid-resolution control on both Pattern grids (#479)

### LINKED ISSUE #479: Pattern grid: ×2 / ÷2 resolution control (ratio-preserving 'more/fewer slots') for Sequencer + Piano Roll
## Problem

A flat Strudel sequence conflates **length** and **resolution**: the token count *is* the number of equal cycle divisions. So `s("bd ~ sn ~ bd")` is exactly 5 slots, and there is no way in the grid to get "more slots without moving the notes" — appending a token re-times everything.

Grounded with real haps:

| pattern | note positions | ratio |
|---|---|---|
| `s("bd ~ sn ~ bd")` (5) | bd@0.0, sn@0.40, bd@0.80 | baseline |
| `s("bd ~ sn ~ bd ~")` (+1) | bd@0.0, sn@0.33, bd@0.67 | **changed** (re-timed) |
| `s("bd ~ ~ ~ sn ~ ~ ~ bd ~")` (×2) | bd@0.0, sn@0.40, bd@0.80 | **preserved** |

Real step sequencers (Logic Step Sequencer) decouple **Rate** (resolution) from **Length** (steps that play). The only ratio-preserving way to change slot count is an **integer resolution change** — arbitrary N→M (e.g. 5→8) lands notes off-grid (polyrhythm) and isn't offered by any mainstream DAW.

## Proposal

A **×2 / ÷2 resolution control** on the Sequencer and Piano Roll headers — the grounded, standard, ratio-preserving form of "more / fewer slots":

- **×2**: each existing slot expands to two; every note keeps its position, the empty slots between hits become editable. `s("bd ~ sn ~ bd")` → `s("bd ~ ~ ~ sn ~ ~ ~ bd ~")`.
- **÷2**: inverse, only enabled when every odd slot is empty (lossless) — otherwise disabled (would drop notes).

This is **pure mini-notation sugar** — the haps are byte-for-byte identical timing, so no semantic change, write-back is the existing pure string surgery (PV123). Shared `tokenize` path covers both grids. Piano Roll already has `@n` holds, which compose cleanly (a held note's span scales with the grid).

## Out of scope (separate, bigger)

- **Add / remove a single trailing step** (change length) — simple but re-times the pattern in a flat sequence; if built, the UI must be honest that it changes the feel, not "extend at fixed rate".
- **Logic-style fixed-rate + adjustable length** = polymeter (`{bd ~ sn}%8`) — the grid currently rejects `{…}` (falls to standby); a genuinely larger feature.

## Acceptance

- ×2/÷2 buttons on both grid headers; ×2 always available, ÷2 enabled only when lossless.
- Round-trip: ×2 then ÷2 returns the byte-identical source.
- Real-hap test: positions before/after ×2 are identical (ratio preserved).
- Unsupported shapes (`@`, foreign gain alignment, etc.) handled or disabled, never corrupting the source.

Deferred — filed for the backlog, not scheduled.

MrityunjayBhardwaj:
Implements #479. The substrate-honest "more / fewer slots" control for both Pattern grids.

## Why this shape (grounded, not a free-form extender)

A flat Strudel sequence conflates **length** and **resolution** — the token count *is* the number of equal cycle subdivisions, so `s("bd ~ sn ~ bd")` is exactly 5 slots and appending a token re-times every event. The only ratio-preserving way to change slot count is an **integer resolution change** (the standard step-sequencer Rate control, decoupled from length). Grounded against real `@strudel` haps:

| pattern | onsets |
|---|---|
| `s("bd ~ sn ~ bd")` | `0, 0.4, 0.8` |
| its ×2 `bd ~ ~ ~ sn ~ ~ ~ bd ~` | `0, 0.4, 0.8` ✅ |
| `note("c3 e3 g3")` | `0, ⅓, ⅔` |
| its ×2 `c3@2 e3@2 g3@2` | `0, ⅓, ⅔` ✅ |
| multi-bar `<bd sn>` → `<[bd ~] [sn ~]>` | `0, 1` ✅ |

## What it does

A **"Slots" ×2 / ÷2** control on the Sequencer and Piano Roll headers:

- **×2** — splits each column in two; every hit keeps its position, the new in-between slots open up. Roll notes scale start **and** duration, so a held `@n` keeps its time span.
- **÷2** — merges pairs back, **disabled** when halving would be lossy (an odd column carries a hit/note, or an odd column count) — the same honest-control rule as the Snap picker: never a silent no-op or corruption.

Pure ratio-preserving mini-notation sugar — the haps are byte-identical before/after, so this is an editing-grid affordance, not a musical change.

## Design

- `notation/resolution.ts` — pure model transform (in the `lane.ts`/`place.ts` family). Returns the **same model ref** when a direction can't apply, so `useGridModel.mutate` skips the write and the document is untouched. Velocity (`.gain`) scales with the grid.
- `ResolutionControl.tsx` — shared control rendered by both grids (Sequencer header row; Piano Roll pinned overlay, so it adds no row and doesn't push the pitch-dense rows — PV143/P216).
- No `PatternPanel` prop threading needed — it's internal via `mutate`.

## Test plan

- [x] editor tsc — 68 (0 new, baseline)
- [x] editor vitest — **2489** (22 new resolution unit tests: ×2/÷2, lossless-guard, round-trip, velocity-scaling, cap, multi-lane)
- [x] app vitest — 571 (unchanged)
- [x] new e2e `resolution.spec.ts` — 3/3 (×2 column-double + write-back, ÷2 disabled-guard + lossless merge, ×2→÷2 byte-identical round-trip)
- [x] **regression**: all 33 grid drag/visual e2e green (piano-roll, sequencer, copy-paste, velocity, snap-division, grid-visual-identity) — the new header/overlay didn't break any drag
- [x] browser-observed: control renders cleanly in both grids, ÷2 correctly dimmed on odd-column patterns
- [x] self-review: multi-bar `<...>` ×2 hap-verified (onsets preserved) — no gating needed

## Stacking

Based on `feat/528-copy-paste-notes` (PR #529) since #479 builds on the post-#527 direct-edit grids. Merge order: **#529 → studio, then #479 → studio** (retarget this PR to `studio_v0.2.0` after #529 lands).

Closes #479


MrityunjayBhardwaj:
Updated per review: the control now presents as **absolute 4 / 8 / 16 / 32 "Slots"** instead of relative ×2/÷2 — same ratio-preserving transform underneath (clicking a target = the right number of lossless ×2/÷2). A preset is enabled only when it's a power-of-2 ratio of the current count and lossless going down; the current count shows active; non-power-of-2 patterns (triplets, hand-written melodies) show every preset disabled rather than re-time.

Commits `17108fb` (refactor src+tests+e2e) + `52fd112` (dist). Gates re-green: editor tsc 68, editor vitest **2498** (31 resolution tests), app vitest 571, resolution e2e 4/4, all 33 grid drag/visual e2e, browser-observed both grids.

Real fixed-rate step **length** editing (the decoupled "number of steps" knob, needs polymeter) is deferred to **#538**.

MrityunjayBhardwaj:
Updated per follow-up: the control now **sets** the grid to N slots, and **quantizes** when the ratio isn't a power of 2 — so *any* pattern can be reduced, including a 64-step choir part (added **64** to the list).

- Lossless (power-of-2 ratio) → same ×2/÷2 as before, hits keep position.
- Otherwise → quantize: notes snap to the nearest new slot, collisions merge (step buckets OR; roll columns become chords), durations clamped so it always serializes (no silent dropped write).
- Quantize targets are drawn dimmer + italic with a `~` prefix and a 'changes timing' tooltip; current count active; multi-bar keeps the lossless-only path.

Verified the 64-step choir reduce against real `@strudel` haps (onsets land on the 32/16/8 grid; collisions stack as chords). Commits `88969a4` + `1bc164c`. Gates: editor tsc 68, editor vitest **2506**, app vitest 571, resolution e2e 5/5 (incl. browser 64→16 reduce), all 33 grid drag/visual e2e, screenshots of both grids.

(#538 — true fixed-rate step *length* via polymeter — remains a separate deferred item; quantize-reduce is lossy coarsening, not fixed-rate extension.)

Issue #601, opened 2026-06-27
Pattern tab: move the slots [4,8,16,32,64] control into the inspector


MrityunjayBhardwaj:
Relocate the slot-count presets from the grid header into the Pattern-tab inspector panel.

MrityunjayBhardwaj:
Landed in studio_v0.2.0 via PR #621 (merge ed52bcec). The Slots [4 8 16 32 64] control now lives in the Pattern inspector instead of the grid headers; each grid lifts its resolution state up via useLiftResolution while still owning the model + quantize write-back. (Closing manually — merge base is studio, not main.)

Issue #477, opened 2026-06-18
Stale e2e: musical-timeline.spec.ts asserts a superseded per-sound row model (4 tests red)


MrityunjayBhardwaj:
## What

During the end-to-end validation pass (session 20), 4 tests in `packages/app/tests/musical-timeline.spec.ts` fail **in isolation** (not flake):

- `s("bd hh cp bd") renders 3 track rows + 4 note blocks`
- `stable track order across re-evals`
- `vocabulary regression on populated live DOM`
- `Phase 20-12.1 — transport stop clears the slot map (drops ghost row)`

## Why (observed in the real browser, not inferred)

A direct DOM probe of the live **Timeline** drawer tab for `s("bd hh cp bd")` shows:

- **rows = `["d1"]`** — one track row keyed by orbit/`$:`-slot, **not** three per-sound rows (`bd`/`hh`/`cp`).
- **8 note blocks** evenly tiled at `left` 0 → 792px (4 per cycle × **2** cycles), each `width` 113px — i.e. a genuine 2-cycle window, not in-place duplication.

This matches the **current, intended** model:

- `timeAxis.ts` → `WINDOW_CYCLES = 2`, and the spec's own *ruler* test (`cycle labels 0/1/2`) passes — confirming the 2-cycle window is by design.
- `MusicalTimeline.tsx` keys rows by track/`$:`-slot (`slotKey`), and renders labels like `d1 · saw` (orbit + optional voice). Sounds are split **inside** a track by the Sequencer (one lane per sound — see the green `sequencer.spec.ts`); voices split on **expand** (#424). So one `$:` line = one timeline row is coherent across the whole milestone.

The failing assertions encode the **old Phase-20 per-sound row model + 1-cycle window**, which the clip-op work this milestone (#442/#444/#446/#463/#472, all of which edited `MusicalTimeline.tsx`) superseded without updating this spec.

## Severity

**Test-debt, not a product regression.** The live timeline behaves correctly; the assertions are stale. App code is byte-identical between `studio_v0.2.0` and the current branch, so this has been red on the integration branch (these are Playwright e2e, not part of the vitest count).

## Fix

Update the 4 assertions to the per-track (`d1`) + 2-cycle-window contract:
- expect 1 track row (`d1`) for a single `$:` line; assert per-sound splitting in `sequencer.spec.ts` instead.
- expect `WINDOW_CYCLES × eventsPerCycle` note blocks.
[...]

PR #622, opened 2026-06-28
✨ Move Note Color from the Pattern tab to Editor Settings (#602)

### LINKED ISSUE #602: Move the note-color [off/velocity] setting from the Pattern tab to the editor settings tab
Note-color toggle is a global preference, not a per-pattern control — move it out of the Pattern tab into the editor Settings tab.

MrityunjayBhardwaj:
## What
Moves the **Note Color** `[Off / Velocity]` control out of the two Pattern grid headers and into the app's **Editor Settings** modal (a "Note color" row under Theme). It's a global editor preference, not a per-pattern control.

Closes #602.

> **Stacked on #601** (`feat/w2-601-slots-inspector`) — both PRs edit the same grid-header regions, so this is stacked to avoid a conflict. Review/merge #621 first; I'll retarget this to `studio_v0.2.0` once it lands. The diff here is #602-only.

## How — the editor↔app same-document seam
The note-color store (`noteColor.ts`) stays the single source of truth. It already used an in-process listener set + `useSyncExternalStore`; #602 just exposes `get/set/subscribeNoteColorMode` through the editor barrel. The Settings modal (a **different package, same document**) calls `setNoteColorMode`, which notifies the listener set, so the **live Pattern grids recolour immediately** — a plain `storage` event wouldn't fire same-document, the shared in-process store does. The grids keep *reading* the mode via `useNoteColorMode`; only the toggle UI moved.

- `NoteColorToggle` is now unused → **deleted**. Both grid headers are simpler; the piano roll drops its top-right overlay and keeps full height for pitch rows.

## Test plan
- editor tsc **68 (0-new)** · app tsc **clean** · dist rebuilt + grep-gated (`get/set/subscribeNoteColorMode` present; `data-note-color-mode` gone; `useNoteColorMode` readers retained).
- e2e: `grid-visual-identity` **3/3** — rewritten to set the mode via **File ▸ Editor Settings…**, asserting the toggle is **gone** from the grid AND that choosing Velocity recolours the **live** cells (proves the cross-package sync). Full Pattern/mixer regression batch **84/84** (grid-visual-identity, resolution, piano-roll, sequencer, pattern-grid-coverage, pattern-horizontal-strip, mixer, mixer-strips, track-identity-pattern).
- Observed live (screenshot): the "Note color" row renders under Theme in the Settings modal.

## Note
`wave-delta-gate` FIXTURE A/B2 fail, but they fail **identically on the base branch** with zero note-color changes — pre-existing flaky/stale timeline specs (#477 family), not introduced here.


MrityunjayBhardwaj:
Landed in studio_v0.2.0 via merge commit 83a4f9a8 (stacked on #621, which merged as ed52bcec). The head commits are already in studio, so GitHub can't auto-mark this MERGED — closing manually. Net diff is in studio.

PR #722, opened 2026-07-04
🎨 Set-bg & backdrop polish: colour bar, focus frame, popover alignment

### LINKED ISSUE #721: Left track colour bar (code view) vanishes when a background viz is set
## Friction

With a background viz set (`data-stave-backdrop="on"`), the per-track colour bar down the editor's left edge (#608) disappears.

## Root cause

The backdrop-transparency blanket in `globals.css` (~L178) sets `background-color: transparent !important` on every `.monaco-editor *`, with an allowlist of `:not(...)` exceptions (selection, caret, suggest/hover/find widgets). The colour-bar overlay segments (`[data-track-colour-bar]`, appended to the editor DOM node) match `.monaco-editor *`, so the `!important` rule overrides their inline `background:<color>` → transparent.

**Observed** (:3000): backdrop off → segments `rgb(249,115,22)`, `rgb(6,182,212)`; force backdrop on → all `rgba(0,0,0,0)`.

## Fix

Add `:not([data-track-colour-bar])` to the allowlist chain so the stripe keeps its identity colour over the viz — the whole point of the bar. App-only (`globals.css`), no editor rebuild.

### LINKED ISSUE #723: Blue focus outline frames the editor when a background viz is active
## Friction

Setting a background viz draws a blue rectangle border around the whole code view.

## Root cause

It's Monaco's editor focus outline — `.monaco-editor` computes `outline: 1px solid rgb(0,127,212)` (`--vscode-focusBorder`). It's present whenever the editor is focused, but invisible against the opaque dark editor surface. With a backdrop on, the editor surface goes transparent (to reveal the viz), so the 1px focus outline becomes a visible blue frame.

**Observed** (:3000, real set-bg flow): the only full-size framed element is `.monaco-editor` with `outline: 1px rgb(0,127,212)`, 1116×751.

## Fix

Suppress the editor outline when the backdrop is active — `[data-stave-code-panel][data-stave-backdrop="on"] .monaco-editor { outline: none }`. Scoped to backdrop mode so normal (non-backdrop) focus indication is unchanged. App-only (`globals.css`).

### LINKED ISSUE #724: Set-bg dropdown popover should left-align to the button, not right-align
## Friction

The "set bg" dropdown (BackdropPopover) on the pattern chrome bar opens *right-aligned* to the button — it extends leftward, so its left edge sits well left of the button. Expected: the popover's **left edge aligns to the button's left edge** (opens down-and-right, the conventional dropdown direction).

## Fix

`BackdropPopover.tsx` positions with `left = anchorRect.right - 320` (right-align). Change to `left = anchorRect.left` (left-align), keeping the viewport clamps so it never overflows. Mirrors `TrackSwatchPopover.tsx`, which already left-aligns via `anchorRect.left`. App-only.

MrityunjayBhardwaj:
Closes #721, closes #723, closes #724. Three small set-bg / backdrop UX fixes in the code view (app-only).

## 1 — Track colour bar vanished over a background viz (#721)
The backdrop transparency blanket (`.monaco-editor * { background-color: transparent !important }`) was blanking the per-track colour-bar segments. Added `:not([data-track-colour-bar])` to its allowlist.

## 2 — Blue focus outline framed the editor under a backdrop (#723)
Monaco's editor focus outline (`.monaco-editor` → `outline: 1px solid rgb(0,127,212)`) reads as a blue rectangle once the editor surface goes transparent for the viz. Suppressed it in backdrop mode (`[data-stave-backdrop="on"] .monaco-editor { outline: none }`); normal focus indication unchanged.

## 3 — Set-bg popover left-aligns to its button (#724)
The dropdown opened right-aligned (`left = anchorRect.right - 320`), extending leftward. Changed to `left = anchorRect.left` so it opens down-and-right, left edge aligned to the button. Viewport clamps kept.

## Diagnosis note
My first pass on #721 verified with a *forced* `data-stave-backdrop` attribute rather than the real "set bg" flow, and reported against stale HMR CSS — so it looked fixed when it wasn't live. All three are now verified through the **real dropdown flow** and each regression test is proven **fail-without / pass-with** where applicable.

## Test plan
- `editor-track-colour-bar.spec.ts` (#721): real set-bg flow, segments stay opaque + byte-identical colour.
- `set-bg-on-pattern-bar.spec.ts` (#723): editor outline is the focus border normally, `none` once a backdrop is pinned.
- `set-bg-on-pattern-bar.spec.ts` (#724): popover left edge aligns to the button (±1px) and extends right.
- Observed all three live on :3000 (screenshots): colour bars visible over the viz; no blue frame; popover opens down-and-right (left 396 == button left 396). App `tsc` clean.

> A pre-existing test in the colour-bar spec ("continuous across a word-wrapped track") fails on clean studio too (±44px word-wrap pixel assertion) — unrelated; flagged for a separate issue.

### DIFF
--- a/packages/app/src/app/globals.css
+++ b/packages/app/src/app/globals.css
@@ -174,9 +174,12 @@ body {
    - the floating popup widgets (suggest / hover / signature help /
      find-replace) — their readable opaque theme surface must survive over
      the viz, else the widget (IntelliSense docs box, or the Cmd+F find
-     widget) goes transparent and unreadable (#598, #681). */
+     widget) goes transparent and unreadable (#598, #681);
+   - the per-track colour-bar segments (`[data-track-colour-bar]`, #608) —
+     the stripe IS its background colour, and showing the track's identity
+     over the viz is the whole point, so blanking it defeats the bar (#721). */
 [data-stave-code-panel][data-stave-backdrop="on"] .monaco-editor,
-[data-stave-code-panel][data-stave-backdrop="on"] .monaco-editor *:not(.selected-text):not(.selectionHighlight):not(.current-line-margin):not(.current-line):not(.minimap-slider-horizontal):not(.cursor):not(.suggest-widget):not(.suggest-widget *):not(.monaco-hover):not(.monaco-hover *):not(.parameter-hints-widget):not(.parameter-hints-widget *):not(.find-widget):not(.find-widget *) {
+[data-stave-code-panel][data-stave-backdrop="on"] .monaco-editor *:not(.selected-text):not(.selectionHighlight):not(.current-line-margin):not(.current-line):not(.minimap-slider-horizontal):not(.cursor):not(.suggest-widget):not(.suggest-widget *):not(.monaco-hover):not(.monaco-hover *):not(.parameter-hints-widget):not(.parameter-hints-widget *):not(.find-widget):not(.find-widget *):not([data-track-colour-bar]) {
   background-color: transparent !important;
 }
 
--- a/packages/app/src/app/globals.css
+++ b/packages/app/src/app/globals.css
@@ -183,6 +183,16 @@ body {
   background-color: transparent !important;
 }
 
+/* Monaco paints a 1px focus outline (`--vscode-focusBorder`, ~#007fd4) on
+   `.monaco-editor` while focused. It's invisible against the opaque editor
+   surface, but with a backdrop on that surface goes transparent and the
+   outline reads as a blue frame around the whole code view (#723). Drop it
+   in backdrop mode; the caret still signals focus, and normal (non-backdrop)
+   focus indication is untouched. */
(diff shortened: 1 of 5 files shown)


MrityunjayBhardwaj:
Added a viz-file regression test (`set-bg-viz-file-outline.spec.ts`) for the #723 outline fix.

A user hit the blue focus frame on a **viz file** (`.p5` / `.hydra`) set as its *own* background. Traced it: the viz-file code panel flips to `data-stave-backdrop="on"` just like the pattern editor, so its `.monaco-editor` paints the same `#007fd4` focus outline (`solid 1px rgb(0,127,212)`), and the existing `[data-stave-code-panel][data-stave-backdrop="on"] .monaco-editor { outline: none !important }` rule already neutralises it — the fix was complete, only the test coverage wasn't. The new test drives the real gesture (open `.p5` → click viz-chrome bg toggle → re-focus) and asserts the computed outline is present before and `none` after. Passes on this branch.

PR #814, opened 2026-07-08
🐛 fix(mixer): instrument picker shows the full live catalog before Play

### LINKED ISSUE #813: 🐛 Instrument picker shows only the curated fallback until (and sometimes after) you press Play
## Problem

Opening the melodic instrument picker (Pattern-tab inspector) shows only the **~27 curated fallback** instruments instead of the full live catalog (~494). Two compounding causes, both confirmed by observing `globalThis.soundMap` in the running app (`undefined` on load → 1853 keys only after a first Play):

1. **The data source is lazy.** The picker reads superdough's global `soundMap`, which the engine only publishes inside `init()` — and `init()` fires on the first **Play/eval**, never on mount. Until you play, `soundMap` doesn't exist, so `MixerBody` renders `liveInstruments ?? INSTRUMENTS` → the curated fallback.

2. **The bridge poll kills itself at zero.** The 1.5s poll in `StaveApp` (`useEffect` around the soundMap accessor) stops after 4 *unchanged* ticks — but 'unchanged' counts being stuck at **0** (engine not yet inited). If you don't Play within ~6s of page load, the poll declares the count 'settled' and clears itself. When you *do* Play later, `soundMap` populates but nothing re-notifies the picker → it stays stuck on the fallback until a remount.

## Fix

- **Poll:** only begin the settle countdown once the count is non-zero; keep polling while the engine is still uninitialized, with a tick ceiling so it never polls forever.
- **Eager warm-up:** on idle after mount, create the active (or first) Strudel file's runtime and call `runtime.init()` (idempotent, no playback, no audio gesture — init just registers sounds + fetches sample manifests) so the full catalog is ready before the user presses Play.

## Verify
Load the app, do NOT press Play, wait for idle, open the instrument picker → full ~494 list. Also: load, wait >10s, then Play → picker fills (poll no longer dead).

MrityunjayBhardwaj:
## Problem

Opening the melodic instrument picker (Pattern-tab inspector) showed only the **~27-item curated fallback** instead of the **~494 live instruments**. Two compounding causes, both confirmed by observing `globalThis.soundMap` in the running app (`undefined` on load → 1853 keys only after a first Play):

1. **The data source is lazy.** The picker reads superdough's global `soundMap`, published only inside `engine.init()` — and init fires on the first **Play/eval**, never on mount. Until you play, `soundMap` is absent, so `MixerBody` renders `liveInstruments ?? INSTRUMENTS` → the fallback.
2. **The bridge poll killed itself at zero.** `StaveApp`'s 1.5s poll stopped after 4 *unchanged* ticks — but 'unchanged' counted being stuck at **0** (engine not yet inited). If you didn't Play within ~6s of load, the poll settled at 0 and cleared itself; a later Play then populated `soundMap` but nothing re-notified the picker → stuck on the fallback until a remount.

## Fix

- **Poll** (`StaveApp.tsx`): only begin the settle countdown once the count is non-zero; keep polling while the engine is still uninitialized, with a tick ceiling (~2 min) so it never polls forever. A slow/late init now refreshes the picker instead of finding a dead poll.
- **Eager warm-up** (`StrudelEditorClient.tsx`): on idle after mount, create the active (or first) Strudel file's runtime and `init()` it — init only registers sounds and fetches sample manifests (**no** `scheduler.start()`, the AudioContext is created *suspended*, so no user gesture and no audio). Runtime + init are idempotent, so a later Play reuses the warmed runtime; `soundMap` is a superdough module singleton, so warming any one engine fills every picker. Retries via the file-list subscription since files load async.

## Test plan

- **Observed (real browser, no Play, no keypress):** fresh page → `soundMap` reaches **1853 keys within ~1.5s** and the picker renders **495 options** across **Synths / Soundfonts / Wavetables / Samples**, sawtooth + soundfonts present, **zero page errors**.
- App unit suite: **630/630 green**.
- No new lint errors/warnings in the two changed files (the one `activeFileIdRef` immutability error is pre-existing on `studio_v0.2.0`).

Closes #813

### DIFF
--- a/packages/app/src/components/StaveApp.tsx
+++ b/packages/app/src/components/StaveApp.tsx
@@ -709,20 +709,34 @@ export function StaveApp({ initialProject }: StaveAppProps) {
     setSoundCatalogAccessor(() => groupSoundCatalog(readDict()));
     let lastCount = -1;
     let stable = 0;
+    let ticks = 0;
+    // Ceiling so we never poll forever if the engine never initializes (eager
+    // warm-up failed AND the user never pressed Play). ~2 min at 1.5s cadence.
+    const MAX_TICKS = 80;
     const tick = () => {
+      ticks += 1;
       const dict = readDict();
       const count = dict ? Object.keys(dict).length : 0;
       if (count !== lastCount) {
         lastCount = count;
         stable = 0;
         notifySoundCatalogChanged();
-      } else {
+      } else if (count > 0) {
+        // Only settle once we actually have a live list. A count stuck at 0
+        // means the engine hasn't inited yet (init is lazy — first Play or the
+        // eager warm-up in StrudelEditorClient) — keep polling so a later init
+        // refreshes the picker instead of leaving it on the curated fallback
+        // (#813). The old `else` incremented here even at 0, so a slow/late
+        // init would find the poll already dead and never repopulate.
         stable += 1;
       }
     };
     const id = window.setInterval(() => {
       tick();
-      if (stable >= 4) window.clearInterval(id); // count settled → stop polling
+      // Stop once a real (non-zero) list has settled, or give up at the ceiling.
+      if ((lastCount > 0 && stable >= 4) || ticks >= MAX_TICKS) {
+        window.clearInterval(id);
+      }
     }, 1500);
     tick();
     return () => {
--- a/packages/app/src/components/StrudelEditorClient.tsx
+++ b/packages/app/src/components/StrudelEditorClient.tsx
@@ -1438,6 +1438,45 @@ export default function StrudelEditorClient({
     });
   }, [runtimeStates, onActiveRuntimeStateChange]);
 
+  // #813 — Eager engine warm-up so the instrument picker is fully populated
+  // before the user presses Play. The picker reads superdough's global
+  // `soundMap`, which is published only when an engine initializes — and init
+  // is lazy (first Play). Without this, opening the picker on a fresh page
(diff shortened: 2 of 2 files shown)


MrityunjayBhardwaj:
## Critical self-review

Audited the diff. One real gap surfaced, confirmed by observation:

**Concurrent double-`init()` → filed as #815 (low severity, follow-up).** The eager warm-up's background `engine.init()` can race the user's first Play and run `init()` twice (observed: two `aliasBank: 1208 → 1853` logs). Both converge on the same 1853-key `soundMap` and the picker renders correctly with zero page errors — the cost is a duplicate manifest fetch + a leaked analyser node. It's latent/pre-existing (`init()` was always unguarded against concurrent callers); the warm-up just makes it easy to hit. Correct fix is an in-flight promise guard on `StrudelEngine.init()`, which touches the sensitive engine file + needs a dist rebuild, so it's split out to keep this PR scoped.

**Otherwise clean:**
- Poll ceiling (`MAX_TICKS = 80`, ~2 min) can't run forever; settle only fires with a non-zero list.
- Warm-up is idempotent (runtime + `init()`), best-effort (`.catch`), fires on idle so it doesn't compete with boot (respects the boot-bounded invariant), and retries via the file-list subscription for async IDB loads.
- No Strudel file in the project → `fid` is null → no warm-up (correct; picker isn't relevant).
- StrictMode double-mount → `warmedRef` guards a single warm.
- No new lint errors/warnings in the two changed files; app unit suite 630/630.

Issue #805, opened 2026-07-07
Audition sounds from the instrument picker (hear-before-you-place)


MrityunjayBhardwaj:
## Problem
The instrument picker lets you *select* a sound but not *hear* it first. Discovery across ~1852 sounds (unlocked by the bridge fix) needs audition — the one capability code+autocomplete can't provide.

## Solution
Reuse the existing `superdough` press-and-hold audition primitive (already in `PianoRollGrid.tsx:261-272`, #633): preview a picker row on click/hover — a mid-register note for melodic instruments, the voice itself for drums. Selection still writes legible `.sound()`/`.bank()` to code (round-trips for both audiences).

## Depends on
The live-registry bridge fix (companion issue) so there's a full list worth auditioning.

## Verify
On the live app, trigger audition from a picker row and confirm audio via a seeded AudioContext meter (headless, not by ear).

MrityunjayBhardwaj:
Instrument (melodic) audition shipped in #806 (merged into studio_v0.2.0): audition-on-select + ▶ re-preview via the shared superdough singleton, observed producing real audio. **Remaining scope for this issue: drum-kit audition** — needs `.bank()` voice resolution to preview a kit. Keeping open for that follow-up.

Issue #633, opened 2026-06-29
Pianoroll: audition pitch on piano-key click/drag (play the track's sound)


MrityunjayBhardwaj:
## Problem

The pianoroll's left-side piano keyboard (the `data-roll-key` column, #430) is purely decorative — it highlights on hover but **plays nothing**. There's no way to audition a pitch while editing a melody, which every DAW piano roll offers (click/drag a key → hear it).

## Fix

Make the piano keys **audition** on press and click-drag:

- **press a key** → play that pitch once.
- **drag across keys** (pointer enters another key while pressed) → play each entered key (natural keyboard glissando).
- Use the **bound track's sound** — read `.s` from the cursor chunk (`readChainMethod(chunk, ['sound','s'])`), fall back to Strudel's default when absent.

### Audio path (grounded)

Strudel's single-hap audio function is `superdough(value, when, dur, cps, cycle)`, exported from `@strudel/webaudio` (a direct editor dep) — a module-global singleton sharing the **same** audio graph the engine drives. The panel imports it directly (it stays document-only otherwise; `superdough` is a stateless audio primitive, not the engine instance). On press: `getAudioContext().resume()` (the click is the unlock gesture) then `superdough({ s, note: midiToPitch(midi), gain }, ctx.currentTime + 0.05, dur, 0.5, 0)`, wrapped in try/catch so audio-not-ready never breaks the UI.

**Known caveat:** synth sounds (`sawtooth`, `sine`, …) always play; sample/soundfont sounds (`gm_*`) need the engine initialized (a played doc) or they no-op silently.

**Scope:** note-name rows only (the piano keys); numeric rows show a value label, not a key.

MrityunjayBhardwaj:
Landed on studio_v0.2.0 (merge ae26767a). Piano-key audition: press plays the track sound, hold sustains, drag glissandos. User-confirmed by ear; e2e guards green.

PR #817, opened 2026-07-08
🐛 fix(audition): ▶ preview sounds identical to a pianoroll key press

### LINKED ISSUE #816: 🐛 Mixer ▶ audition sounds different from a pianoroll key press (envelope drift)
## Problem

Auditioning an instrument two ways produces two different sounds even though both use the **same instrument** (both read the chunk's `.sound()` via `readChainMethod(chunk, ['sound','s'])` — picker at MixerBody.tsx:406, pianoroll at PianoRollGrid.tsx:263).

The difference is the **envelope**: `auditionSound` (audition.ts, extracted in #805 from PianoRollGrid's #633 press-and-hold) drifted from `playMidi`:

| param | pianoroll `playMidi` | mixer ▶ `auditionSound` |
|---|---|---|
| sustain | 1.0 | 0.5 |
| release | 0.08 | 0.12 |
| note length | 0.22s (HOLD_RETRIGGER_MS) | 0.5s |

So the ▶ preview is softer (half sustain) and rings ~2.3× longer than a key tap — same note, different articulation. Edge case: a chunk with no `.sound()` plays the default synth on a key press but disables ▶, so they can't be A/B'd there.

## Fix

Extract a single shared audition envelope + note-length constant (in audition.ts) used by BOTH `auditionSound` and `playMidi`, so a ▶ preview sounds identical to a key tap and the two can never drift again. Derive PianoRollGrid's HOLD_RETRIGGER_MS from the shared duration.

## Verify

Trigger both paths on the same note/instrument, capture the `value` object superdough receives — sustain/release/duration must match.

MrityunjayBhardwaj:
## Problem

Auditioning an instrument two ways produced two different sounds even though both use the **same instrument** (both read the chunk's `.sound()` via `readChainMethod(chunk, ['sound','s'])` — picker MixerBody.tsx:406, pianoroll PianoRollGrid.tsx:263).

The difference was the **envelope**: `auditionSound` (audition.ts, extracted in #805 from PianoRollGrid's #633 press-and-hold) drifted from `playMidi`:

| param | pianoroll `playMidi` | mixer ▶ `auditionSound` |
|---|---|---|
| sustain | 1.0 | 0.5 |
| release | 0.08 | 0.12 |
| note length | 0.22s | 0.5s |

→ the ▶ preview was softer (half sustain) and rang ~2.3× longer than a key tap.

## Fix

Extract one shared `AUDITION_ENVELOPE` + `AUDITION_DUR_S` (audition.ts) used by **both** `auditionSound` and `playMidi`, so a ▶ preview and a key tap trigger the same note shape and can't drift again. `HOLD_RETRIGGER_MS` is derived from the shared duration.

## Test plan

- **Observed (real browser):** a pianoroll key press and a ▶ click now emit an **identical gain envelope** (same 5 automation events, same sustain plateau at 1, previously 0.5), zero page errors.
- Regression test locks `auditionSound`'s emitted gain/attack/sustain/release/duration to the shared constants.
- Editor suite **2848/2848**. Rebuilt `@stave/editor` dist.

Closes #816


MrityunjayBhardwaj:
## Critical self-review

- **No third path.** Grepped all `superdough(` audition call sites: only `audition.ts` (auditionSound) and `PianoRollGrid.tsx` (playMidi) — both now use the shared `AUDITION_ENVELOPE`/`AUDITION_DUR_S`. The `sustain:/attack:` literals in `knobRanges.ts` are effect-knob *ranges*, unrelated.
- **Held-key behavior preserved.** `HOLD_RETRIGGER_MS = AUDITION_DUR_S * 1000 = 220` — byte-identical to the old literal, so the press-and-hold retrigger/glissando is unchanged.
- **No-sound edge case (unchanged, by design).** A chunk with no `.sound()` still plays the default synth on a key press while ▶ is disabled (empty value). You can't audition 'nothing' via ▶ — the fix targets the with-sound case the report was about.
- **Duration coupling is intentional.** `AUDITION_DUR_S` doubles as the pianoroll retrigger interval so a tap = a one-shot = 0.22s; if a longer preview is ever wanted, decouple then (documented in the constant's comment).
- Regression test locks the emitted envelope to the shared constants; observed identical gain automation live.

PR #833, opened 2026-07-09
✨ Asset Library: Viz preset provider + default-to-type coexistence (#832)

### LINKED ISSUE #832: Asset Library Phase 2a: Viz preset provider + default-to-type coexistence
### Goal
Make saved viz presets a first-class asset type in the Library, alongside Sounds, and fix how the two coexist in one panel.

### Part 1 — Viz provider (app)
A `createVizProvider` (mirrors `createSoundsProvider`) that maps `VizPreset → Asset`:
- `type: "viz"`, `id: preset.id` (unique row key), `name: preset.name`
- `source`: `isBundledPresetId(id) ? "built-in" : "user"`
- `group`: renderer (`P5` / `Hydra` / `GLSL`) → muted label + category chips
- `tags`: `[renderer, ...requires]`
- **insert** writes `.viz("<preset.name>")` — `.viz()` resolves by NAME, not id (`registerNamedViz(preset.name,…)`), so the code string is the name while the row key is the id. They diverge here (for sounds they're identical).
- **preview**: omitted for the MVP (viz is visual, the 40px row has no thumbnail slot — thumbnail is a deliberate follow-up).

Reads `VizPresetStore.getAll()` (async IndexedDB) into a cache; `list()` is synchronous and the load calls `notifyAssetProvidersChanged()`. Pure mapping is dependency-injected (editor-free) so it unit-tests without IndexedDB — same reason as the Sounds provider.

Because Copy currently copies `asset.id` (== code string for sounds), add an optional `Asset.code` (the insertable/copyable code token, defaults to `id`); viz sets it to `preset.name` so Copy yields a string that actually resolves in `.viz()`.

### Part 2 — Viz cursor-writer (editor)
New `assignVizToCursor(name)` on the shell handle + a pure `planVizAssignment(doc, offset, name)` (sibling of `planSoundAssignment`): append `.viz("name")` to the pattern chunk under the cursor, or replace an existing `.viz()` argument. No chunk under the cursor → no-op (a viz must attach to a pattern; the richer no-chunk case is a follow-up). Unit-tested; editor `dist` rebuilt.

### Part 3 — Coexistence: default to a type, not All
With `types.length > 1`, `AssetLibraryPanel` initializes `type` to the first available type instead of `null` (so the panel lands on one clean taxonomy, not the 1851+12 firehose with mixed category chips). "All" stays clickable. The existing "drop filter if its value vanished" effect must fall back to another available type, not `null`.

Thumbnail preview variants are a separate follow-up.

MrityunjayBhardwaj:
**Stacked on #831** (base `feat/830-sound-assign-planner`) — it builds on that PR's `planSoundAssignment` in `WorkspaceShell`. Retarget to `studio_v0.2.0` once #831 merges.

### What
Makes saved viz presets a first-class asset type in the Library, alongside Sounds, and answers how the two coexist in one panel.

### Part 1 — Viz provider (app)
`createVizProvider` maps each `VizPreset → Asset`:
- `id` = `preset.id` (unique row key); `name` = `preset.name`
- `group` = renderer (`P5` / `Hydra` / `GLSL`) → muted label + category chips
- `source` = built-in (bundled) / user; `tags` = `[renderer, …requires]`
- **insert** writes `.viz("<preset.name>")` — `.viz()` resolves by NAME, not id, so the code token and the row key diverge (for sounds they're the same). A new optional **`Asset.code`** (defaults to `id`) carries that so **Copy** yields a string that actually resolves.
- **No preview** (MVP) — a viz is visual and the 40px row has no thumbnail slot; thumbnail variants are a deliberate follow-up.

Reads `VizPresetStore` (IndexedDB) into a ref cache; `list()` is synchronous, and it refetches when the Library panel opens (no preset-change event to subscribe to). IDB read is bounded (never hangs).

### Part 2 — Viz cursor-writer (editor)
New `assignVizToCursor(name)` on the shell handle, backed by a pure, unit-tested **`planVizAssignment(doc, offset, name)`** (sibling of `planSoundAssignment`): append `.viz("name")` to the pattern chunk under the cursor, or replace an existing `.viz()` arg (preserving any options object). No chunk → no-op (a viz decorates a pattern). Double-quoted, matching the shipped `.viz()` convention.

### Part 3 — Coexistence: default to a type, not All
The panel defaults `type` to the first available type (once >1 exists) instead of the noisy "All" union — sounds outnumber viz ~1851:12, so "All" buries viz and mixes two category taxonomies. "All" stays one click away; the vanished-value guard falls back to another type, not null.

### Test plan
- **New unit tests (16):** `planVizAssignment` (8 — append/replace/options-preserve/escape/no-chunk) and `vizProvider` (8 — id/name divergence, insert writes the name, source, group, sort, isLoading).
- Editor **2883/2883** + app **677/677** green; app `tsc --noEmit` clean; editor `dist` rebuilt + committed.
- **Observed in the real app:** Library lands on Sounds; the Viz chip lists presets grouped `GLSL · Hydra · P5` with counts (`All 18 · P5 11 · Hydra 5 · GLSL 2`); two same-named "Piano Roll" presets render as distinct rows (validating id-as-key); inserting Aurora writes `note("c4 e4 g4").viz("Aurora")`.

Closes #832


MrityunjayBhardwaj:
Landed on studio_v0.2.0 via a clean fast-forward of the viz-library stack (#831 → #833 → #835 → #837). studio isn't the default branch, so closing manually; studio→main will be formalized at the #238 gate.

Issue #842, opened 2026-07-10
Mixer shows phantom dials for a control's extra (Strudel-ignored) args


MrityunjayBhardwaj:
## Problem

In the Mixer, `.room(0.25)` shows one Reverb dial that tracks the value — correct. But `.room(0.25, 0, 100)` renders **three** separate "room" dials (`room 1`, `room 2`, `room 3`).

**Root cause.** The drawer builds one knob per numeric argument of every chain call (`knobsFromChunk` in `MixerBody.tsx`). That's right for genuinely multi-arg functions (`.euclid(3,8)`), but wrong for Strudel *controls*. Grounded in `@strudel/core` `controls.mjs:50`:

```js
Pattern.prototype[name] = function (value) { return func(value, this); };
```

Every control created by `createParam` (room, gain, lpf, delay, crush, …) is a **unary** prototype method — it reads only the first argument. So `.room(0.25, 0, 100)` plays identically to `.room(0.25)`; JavaScript discards the `0` and `100`. Dials 2 and 3 therefore edit literals with **zero runtime effect** — a false projection that breaks the Mixer's "trustworthy projection of the document" invariant.

(Note: the extra args are not a "range" mechanism — Strudel has no such syntax, so writing a range there is silently ignored.)

## Fix

Cap a **known control** to a single dial (its first numeric arg). Genuinely multi-arg functions (not in the control set) keep one knob per argument. Grounded predicate `isKnownControl` over the Mixer's range table (a superset of the effect catalog); all its entries are unary Strudel controls.

MrityunjayBhardwaj:
Fixed in `studio_v0.2.0` via #843 (fast-forwarded). One dial per unary control; extra Strudel-ignored args no longer spawn phantom dials.

PR #845, opened 2026-07-10
✨ feat(mixer): custom dial range via double-click popup, linked to code

### LINKED ISSUE #844: Custom per-dial range via double-click popup, linked to code
## Goal

Give each Mixer dial an editable range. Double-clicking a dial opens a small in-place popup with two inputs (start / end); setting them writes a custom range into the code, and editing the code updates the dial — bidirectionally linked.

## Encoding (decided)

Positional args on a **known single-value control**: `.room(0.25, 0, 100)` → value `0.25`, dial range `[0, 100]`. Grounded: Strudel controls are unary (`@strudel/core` `controls.mjs:50`), so args 2+ are ignored at runtime — the range is Stave metadata that never changes playback. Continues #842 (which caps a control to one dial); this gives its extra args a real Stave meaning without resurrecting phantom dials.

Only control dials are range-editable — genuinely multi-arg functions (`euclid(3,8)`) use their args for real, so they're excluded (gated by the `isKnownControl` set).

## Behavior

- **Read:** `knobsFromChunk` attaches `customRange {min,max}` from args[1]/args[2] when both are numeric; the dial renders with that range instead of the default.
- **Write:** double-click → popup (start/end) → insert `, min, max` after the value arg (or replace the existing span); a "Reset" removes them, returning to `.control(value)`.
- **Bidirectional:** code→UI is automatic (the strip is a pure re-derivation of the text); UI→code goes through the tagged knob writeback (one undo step, live re-eval).

## Test plan

- Read: `.room(0.25, 0, 100)` → dial `customRange {0,100}`.
- Write (pure `rangeArgsEdit`): `.room(0.25)` → `.room(0.25, 0, 100)`; existing span replaced in place; reset removes.
- Round-trip: write then re-parse yields the same range.

MrityunjayBhardwaj:
Stacked on #843 (needs its `isKnownControl` + one-dial-per-control). GitHub will retarget this to `studio_v0.2.0` when #843 merges.

## What

Double-click a Mixer dial → a small in-place popup with **Start / End** inputs. Setting them stores a custom range in code as `.control(value, min, max)`; editing that code updates the dial. Bidirectional.

## Encoding

Positional args on a known single-value control. Strudel controls are unary (`@strudel/core` `controls.mjs:50`), so args 2+ are ignored at runtime — the range is Stave metadata that never changes playback. Gives the extra args a real meaning without bringing back the phantom dials #842 removed. Multi-arg functions (`euclid`, …) aren't controls → not range-editable (their args are real).

## How

- **Read:** `knobsFromChunk` attaches `customRange {min,max}` from `args[1]/args[2]`; the dial renders with it instead of the default.
- **Write:** pure `rangeArgsEdit` / `rangeResetEdit` insert / replace / remove the range slots; the popup routes them through the tagged knob writeback (one undo, live re-eval). Code→UI is the existing strip re-derivation — automatic.
- **Reset** in the popup clears the range back to `.control(value)`.

## Test plan

- Read + write + round-trip: `.room(0.25, 0, 100)` → `{0,100}`; insert / replace / normalise-lone-arg / reset / no-op (`knobsFromChunk.test.ts`).
- Popup DOM interaction: double-click opens prefilled, Apply reports the range, zero-span ignored, Reset fires, non-editable dial does nothing (`Knob.range.test.tsx`).
- `visualEdit` suite green: **594 passed**. Dist rebuilt with src.

closes #844


MrityunjayBhardwaj:
Landed in `studio_v0.2.0` via fast-forward (commits e849d0f5 + 929078da), merged together with its stacked parent #843 to avoid a minified-dist three-way conflict. Closing since the changes are already on the branch — GitHub can't flip this to "merged" because its base was the now-removed parent branch.

Issue #875, opened 2026-07-12
Red Playwright gate: 41 genuine failures in 2 clusters + 12 singles


MrityunjayBhardwaj:
Triage of the full app Playwright suite on clean `studio_v0.2.0` (80ba0236): **105 failed / 358 passed**.

Verified all of these fail **in isolation** as well as in the suite, and reproduce identically on the pre-#870/#873 base (cb1cabdc) — none are caused by the recent eval-backed timeline work.

## Breakdown

- **49 — IR Inspector specs** (`ir-inspector-{tier4,timeline,debugger,mode-toggle}`). The panel's tab registration is deliberately commented out (`StaveApp.tsx`, "re-enable: #680"), so the UI is unreachable and the specs time out. Quarantined against #680 in the cleanup PR.
- **15 — specs for the retired DOM live-view** (`musical-timeline*`, `click-to-source-test`). Retired / quarantined; the capability question in click-to-source is tracked separately.
- **41 — GENUINE, still open.** This issue.

## The 41

**Cluster A — canvas context conflict (15).** `strudel-viz-methods` (13 of 26) + `viz-single-source` (2):
```
InvalidStateError: Failed to execute 'getContext' on 'HTMLCanvasElement':
Cannot get context from a canvas that has already been [used with a different context type]
```
One root cause, two files. Smells like a real bug (a canvas re-acquired with a different context type), not a stale selector.

**Cluster B — lane keys / labels (~14).** `track-rename` (5), `track-custom-color` (4), `timeline-caret-lane-select` (2), `wave-delta-gate` (2), `full-song-track-labels` (1). All revolve around lane keys/labels (`d1`/`d2` vs named tracks). The loudest is `track-rename`, which fails on a React **"Encountered two children with the same key"** console error — a genuine defect, not a spec problem. Plausibly one root cause across several files.

**Singles (12).** `full-song-timeline:93` (count settle race), `viz-chain`, `hydra-stave-bag` ("no hydra tab found in default project"), `viz-library`, `visual-edit-tabs`, `perf-profiler`, `mixer-dial-range`, `grid-visual-identity` (strict-mode violation — two "File" buttons now match), `editor-track-colour-bar`, `backdrop-viz-chrome`.

## Also

`npx playwright test` cannot run as one command: `testDir: ./tests` has no `testIgnore`, so it sweeps in the vitest-only `parity-corpus/*.spec.ts` and dies on `require is not defined`. Fixed in the cleanup PR.

[...]

Issue #874, opened 2026-07-12
Note-level click-to-source may have been lost with the DOM live-view retirement


MrityunjayBhardwaj:
Found while triaging the red Playwright gate.

`click-to-source-test.spec.ts` clicks an individual **note block** in the timeline and asserts the editor cursor lands on that note's source line. It drives `[data-musical-timeline-note="..."]` — DOM that the #497/U5 live-view retirement deleted. The spec has been failing ever since.

The canvas timeline has **lane-header** jump (#610: click a lane header → reveal that track's code) but, as far as I can tell, no **note-level** jump. So the finer capability — click *this note*, land on *its* line — appears to have quietly disappeared with the DOM view rather than being deliberately dropped.

Two possible resolutions:
1. It was intentional (lane-level jump is enough) → retire the spec.
2. It's a capability regression → restore note-level click-to-source on the canvas (hit-test the mark → `loc` → cursor), and rewrite the spec against the canvas.

The spec is quarantined (skipped, pointing here) in the gate-cleanup PR so it stops masking the rest of the red — it should not stay skipped indefinitely.

MrityunjayBhardwaj:
**Decision: the capability was LOST, not deliberately dropped — restore it.**

Note-level click-to-source should still work on the canvas timeline. Treating its absence as a regression, not an intended narrowing to lane-level jump.

Implication: this needs a hit-test from canvas pixels back to the source token — the note marks already carry their source locations (the eval-backed marks from #862 are built from `hap.context.locations`), so the mapping exists; what's missing is the pointer→mark hit-test on the canvas and the jump.

The spec stays SKIPPED (not deleted) until that lands — it is the acceptance test.

MrityunjayBhardwaj:
## Diagnosed: it vanished — and the data to restore it is already in hand

Answering this issue's own question (deliberate drop vs capability regression) with the record rather than a guess: **it was a regression**, and the plan below is grounded in what the code actually does today.

### Why it disappeared

The retirement commit (`1d787515`, #502/U5) names it in its own summary of what it deleted:

> "a DOM-rendered live 2-cycle window (track rows + note blocks + per-hap glow + **click-to-source**)" … "Removed the entire window renderer: … **handleNoteClick**, …"

on the reasoning that "with the per-note live overlay shipped (#500/U3), **the canvas view subsumes the live monitor**, so the fork was pure redundancy."

[...]

Issue #872, opened 2026-07-12
Flaky timeline specs: setValue races the project file load, so the app evaluates the starter example


MrityunjayBhardwaj:
## Problem

The timeline Playwright specs seed their code with a Monaco `model.setValue(...)` right after `boot()` (which only waits for a Monaco editor to exist), then press Cmd+Enter. But the editor's content is a **controlled value** fed from the project file store (async). When that load lands *after* the spec's `setValue`, it overwrites the test's code and the app evaluates **the starter example** instead.

Observed on `full-song-eval-lanes.spec.ts` (it seeds a 2-track song):

```
Expected: 2   ← its own two tracks
Received: 3   ← the starter example's three lanes
```

It passes in isolation and fails intermittently inside a batch, where boot is slower — which is exactly the ordering the race predicts. It bit twice while working on #863/#871, and it silently invalidated a manual before/after probe (the "before" run measured the starter, not the fixture).

This is worse than an ordinary flake: when it loses the race the spec still *runs*, just against the wrong song. Assertions that happen to hold for the starter would pass green on a fixture that was never evaluated.

## Fix sketch

Wait for the file content to actually land before seeding — the model value goes from empty to the loaded file, so `waitForFunction(() => model.getValue().length > 0)` orders it correctly. Add that to `boot()`/`evalCode()`.

The helper is **copy-pasted across ~40 `full-song-*.spec.ts` files**, so the durable fix is to lift `boot`/`evalCode` into one shared test helper module and guard it there — pairs naturally with the stale-timeline-spec cleanup.

MrityunjayBhardwaj:
PR #888 lands the shared guarded helper (`tests/_appBoot.ts`) + the audit. Summary of what the audit found, since it changes what the rest of this issue is worth.

## The audit

The issue predicted the dangerous case in its own words: *"assertions that happen to hold for the starter would pass green on a fixture that was never evaluated."* So I measured it rather than guessing who that hit.

**Seed-blind run:** a hook lets each spec's seed land, then restores the starter — exactly what a lost race does. A spec that still passes was never testing its own fixture.

**381 seed-blind tests → 313 correctly FAILED.** For those, the race can only make them flaky, never dishonest. That is the reassuring half.

## The suspects, honestly filtered

24 files still passed, but most are **instrument blind spots, not findings**:

[...]

Issue #838, opened 2026-07-09
Live hover thumbnails for viz library packages


MrityunjayBhardwaj:
## Problem
Viz-library cards currently show a **baked** thumbnail — a still frame captured once, offline, through the compositor-capture harness. That works for the two curated GLSL packages, but it doesn't extend to **user-authored** packages: there's no offline bake step for a viz a user just wrote, so their card falls back to the flat renderer-hued placeholder tile.

## Solution — ShaderToy-style live hover
Keep the static tile as the resting state. On **hover**, render the real viz **live** in the ~64px card (in the dedicated viz worker, presented on-screen — no frame readback), driven by a **default, muted** audio-signal source so audio-reactive shaders actually move. Tear down on mouse-leave. Only one preview is ever alive at once, so this stays cheap on the shared GPU.

The "default audio" reuses the **drum-pattern example we already have** — `s("bd*4, ~ sd, hh*8").bank("RolandTR909")`, the exact pattern the thumbnail baker uses and which is already proven to light up both shader classes (Prism's spectrum and Pulse Grid's events). No sound is produced; only the signals drive the visuals.

## Plan (atomic commits)
1. **Demo signal source** — a looping, muted signal generator derived from the drum pattern (kick×4, snare on 2 & 4, hats×8) + a matching pulsing spectrum. Pure, unit-tested: assert the spectrum varies over time and the events fire.
2. **App → panel bridge** — the app injects a `mountVizPreview` capability into the library panel (keeps the panel and its tests free of the heavy editor barrel).
3. **Card hover** — mount on enter / dispose on leave, static tile underneath, first-frame crossfade.
4. **Tests + live observation** — hover Prism → shader animates, reactive, silent; leave → torn down; worker context released.

## Non-goals
- No frame capture / ImageBitmap readback (the worker presents on-screen directly).
- No change to the curated baked-thumbnail path (it remains the resting-state tile).

MrityunjayBhardwaj:
Shipped in #839, fast-forwarded into `studio_v0.2.0` at 83305830.

Viz-library cards now render the real shader live on hover — muted but audio-reactive against a synthetic drum-pattern signal — and tear down on mouse-leave. Also added a Settings → Visualization "Preview height" slider with aspect-true card sizing.

Closing manually since the PR landed on `studio_v0.2.0` (not the default branch), so GitHub didn't auto-close.

PR #884, opened 2026-07-14
🧹 chore(tests): clear the last 14 reds — the Playwright gate goes green (#875)

MrityunjayBhardwaj:
The last stretch of #875. The gate on merged `studio_v0.2.0` was **14 failed / 387 passed / 131 skipped**. It is now **0 failed / 403 passed / 131 skipped** — 534 collected, reconciled (403 + 131 = 534), exit 0.

## The headline

**Zero product bugs.** Every one of the 14 was the instrument, not the instrument's subject. That is worth saying plainly because the failure *messages* said otherwise — "two File buttons", "audio still playing after close", "only 1 lane", "element not found". A red spec's message describes the **assertion**, not the product; ranking these by how alarming they sounded mis-ranked them every time.

The reds also cluster **by cause, not by file**. One structural move rots specs in suites that have nothing to do with each other: the unified settings shell (#739) broke a grid spec and a viz spec; the worker-default (#245) broke canvas probes, `window.__x` readback channels and profiler section names alike.

## What each one actually was

| Spec | What it claimed | What was true |
|---|---|---|
| `viz-chain` | a chain shape broke | `Prism` stopped being seeded when the Asset Library landed (#834) |
| `hydra-stave-bag` | the stave bag is missing | probe scanned open tabs, handed its result back on a `globalThis` that doesn't exist off-main, and seeded the editor by keystrokes |
| `backdrop-viz-chrome` | audio still plays after close | the viz's own context **is** closed; the survivor is the shared superdough context — the oracle **demanded a regression** |
| `editor-track-colour-bar` | the bar is 44px short | `getBottomForLineNumber` is not wrap-aware; the bar renders exactly the wrapped block |
| `perf-profiler` | the profiler yields nothing | it yields plenty, under the worker's vocabulary — `*.bus` is the pre-#245 main-thread name |
| `visual-edit-tabs` | Mixer should be gone | #540 deliberately re-added it as a **peer** console (`tabs.ts:52`) |
| `grid-visual-identity` | there are two File buttons | `getByRole(name:)` is a case-insensitive **substring** match; and Settings has since moved into the shell (#739) |
| `mixer-dial-range` | the range didn't arrive | it did; `.room(0.4, 10, 90)` puts the value outside its own range and #848 deliberately **widens** to keep it representable |
| `viz-library` | the cards aren't side by side | the flex grid reflows; at the #838 default card height two simply don't fit in a 260px sidebar |
| `full-song-timeline` | analysis renders < 2 lanes | a bare `stack(a, b)` is **one** track; it only ever passed because the seed lost a race and the 3-track **starter** got analyzed |
| `mixer` | the menu item is missing | the only genuinely load-dependent red: it clicked ＋Add-effect without waiting for the strip to bind |

## Two vacuous greens closed as a side-effect

- `mixer-dial-range` had no coverage for the widening guardrail's **below-min** direction — which is exactly why nobody caught the incoherent fixture. Added, at both the unit and dial level.
- `viz-density-lod` carried the same latent `File` substring trap. It is env-gated so it never failed the gate; fixed anyway.

## Test plan

- `npx playwright test` — **0 failed / 403 passed / 131 skipped**, 534 collected.
- Editor unit suite green (knob guardrails 10/10, incl. the new below-min + inside-range cases).
- The viz-library reflow is verified by **looking**, not by exit code: `viz-library-thumbs.png` (one column at the default height) and `viz-library-reflow.png` (side by side at the minimum).

## Still open, deliberately

- **#872** — the shared guarded boot helper. 61 specs still wait only for monaco to *exist*. It will not clear reds (there are none), but it is the root of the false green in `full-song-timeline` above, so it is now a proven correctness issue rather than a tidiness one.
- **#874**, **#737** — untouched.

Refs #875.

### DIFF
--- a/packages/app/tests/backdrop-viz-chrome.spec.ts
+++ b/packages/app/tests/backdrop-viz-chrome.spec.ts
@@ -322,14 +322,42 @@ test.describe('Backdrop via viz-settings popover (#773)', () => {
     const originalGroupId = await backdrop
       .first()
       .getAttribute('data-workspace-background')
-    const runningCtxs = () =>
-      page.evaluate(
-        () =>
-          (window as unknown as { __ctxs: AudioContext[] }).__ctxs.filter(
-            (c) => c.state === 'running',
-          ).length,
-      )
-    expect(await runningCtxs()).toBeGreaterThan(0)
+    // Count the running contexts the VIZ owns — i.e. every context EXCEPT the
+    // app's shared one (`getAudioContext()`, superdough's).
+    //
+    // This used to count every running AudioContext and require ZERO after the
+    // close. That is the wrong requirement: sample audition now runs through the
+    // SHARED context, so opening the preview lazily creates it, and an
+    // AudioContext can never be reopened once closed — tearing it down on a tab
+    // close would permanently kill all app audio. Observed: the viz's own context
+    // IS closed on close (state "closed"); the survivor is `getAudioContext()`
+    // itself, still ticking, exactly as it should. The old assertion turned that
+    // correct behaviour into a red (#875).
+    const runningVizCtxs = () =>
+      page.evaluate(() => {
+        const w = window as unknown as {
+          __ctxs: AudioContext[]
+          getAudioContext?: () => AudioContext
+        }
+        let shared: AudioContext | null = null
+        try {
+          shared = w.getAudioContext?.() ?? null
+        } catch {
+          shared = null
+        }
+        return w.__ctxs.filter((c) => c !== shared && c.state === 'running').length
+      })
+    /** The shared context must SURVIVE the close — the opposite regression. */
+    const sharedAlive = () =>
+      page.evaluate(() => {
+        const w = window as unknown as { getAudioContext?: () => AudioContext }
+        try {
+          return w.getAudioContext?.()?.state !== 'closed'
+        } catch {
+          return false
+        }
+      })
+    expect(await runningVizCtxs()).toBeGreaterThan(0)
 
     // Drag the hydra EDITOR tab to the east quadrant of its group → new pane.
     const editorTab = page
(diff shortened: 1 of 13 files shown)


MrityunjayBhardwaj:
## Correction on the gate number — and an honest one

I opened this claiming **0 failed / 403 passed**. That was one run. I ran the gate four times, and it is not stably green:

| run | failed | which |
|---|---|---|
| 1 | 2 | `full-song-timeline:93`, `mixer:232` — both **fixed in this PR** |
| 2 | **0** | — |
| 3 | 3 | `full-song-follow`, `full-song-live-overlay`, `viz-shared-pump-observe` |
| 4 | 1 | `stave:130` (a11y single-H1) |

Every run reconciles against 534 collected.

**The 14 original singles are fixed and none of them recur.** What's left is a *rotating* cast of pre-existing intermittent specs — a different subset each run, none in a file this PR touches:

- `stave:130`, `viz-shared-pump-observe`, `full-song-live-overlay` — pass in isolation, fail only under parallel load.
- `full-song-follow` — the playhead race, now filed as **#885**.

These are **not caused by this branch**: `git diff --name-only cad377cd..HEAD` is test files only — **zero product source** — so the app under test is byte-identical to `studio_v0.2.0`.

## What I deliberately did NOT do

I had a fix in hand for `full-song-follow` (wait for the editor to hold content before evaluating — the #872 guard) and **reverted it**, because observation killed the hypothesis: the editor *does* hold the song, the eval *does* land (3 lanes, 0 console errors), and the playhead still doesn't appear. The guard changed no outcome, so shipping it would have been unverified ballast over a bug I hadn't understood.

What's actually happening is worth its own issue, so it got one: **#885** — after a successful evaluate, the song position intermittently stays null, so the playhead never renders. Same config gives a playhead on one run and none on the next. If that's playback genuinely not starting, it's user-visible (hit play, get no playhead), which is a good deal more interesting than a red spec.

Issue #882, opened 2026-07-13
Editing a viz option's value and re-evaluating clears the old option but never applies the new one


MrityunjayBhardwaj:
## Problem

Change **only** the value of a viz option on a live backdrop and re-evaluate (Cmd+Enter, no stop+play):

```js
$: note("c4 e4 g4 c5").s("sawtooth").pianoroll({ background: '#cc1133' })   // red  ✓
$: note("c4 e4 g4 c5").s("sawtooth").pianoroll({ background: '#11cc33' })   // → paints NEITHER
```

Observed (compositor pixel fractions over the backdrop canvas):

| step | red | green |
|---|---|---|
| after `#cc1133` | **0.700** | 0.000 |
| after editing to `#11cc33`, re-evaluate | **0.000** | **0.000** |

The old option correctly stops applying, but the new value never arrives — the sketch falls back to its default (transparent) background. So the second evaluate delivers an EMPTY options bag rather than the edited one.

## Not the worker boundary

This reproduces **identically on the main-thread renderer** (`stave.viz.worker=0`), so it is not #880 (viz options not crossing the worker boundary) and is not fixed by PR #881 — it sits further upstream, in how the engine captures `backdropVizOptions` on a re-evaluate.

Suspected: the backdrop option capture (`capturedBackdropVizOptions` in `StrudelEngine.evaluate`) only survives the first evaluate; the re-eval publishes a backdrop request whose `options` are empty, so the renderer dutifully applies `{}`.

## Why it was never caught

Every spec that changes an option does a **stop+play** (`Cmd+.` then re-run) between the two states — which re-mounts and re-captures. The plain re-evaluate gesture (what a user actually does while tweaking a colour) is uncovered.

## Repro

1. Run `$: note("c4 e4 g4 c5").s("sawtooth").pianoroll({ background: '#cc1133' })` → backdrop is red.
2. Edit the hex to `#11cc33`. Press Cmd+Enter (do NOT stop).
3. Backdrop turns transparent instead of green.

Note the same "options must reach the sketch" assertions now exist and are falsifiable (PR #881), so a fix here has a working oracle to land against.

MrityunjayBhardwaj:
## Closing — observation refutes the premise: viz options are fine

Went to fix this and it collapsed on contact. **Viz options are not dropped on re-evaluate.** The reported symptom is real, but the cause is the *gesture in the repro*, not the options plumbing.

[...]

Issue #180, opened 2026-05-27
Ctrl+Enter while playing is a silent no-op — re-eval requires stop+play (or live mode)


MrityunjayBhardwaj:
**Observation surfaced during #174 / PR #178 testing.** With audio playing, pressing Ctrl+Enter again does NOT re-evaluate. `onPlay` → `runtime.play()` doesn't appear to fire `fireEvaluateSuccess()` on the second call, so any hook on `onEvaluateSuccess` (including the code-driven backdrop hook from #174) only sees the *first* play. Real re-eval happens via:
- **Live mode** (auto-refresh on) — file edits trigger `onLiveModeContentChanged` → `play()` → fires success
- **Manual stop+play** (Ctrl+. then Ctrl+Enter)

If the intended UX is "Ctrl+Enter re-evaluates while playing" (common live-coding convention), it's silently broken. If "Ctrl+Enter is play, edit live or stop first to re-eval" is the intended model, that should be discoverable (currently silent). Worth a quick investigation in `runtime/strudelRuntime.tsx` + `EditorView.tsx` Ctrl+Enter command path to determine intent vs bug.

MrityunjayBhardwaj:
## Diagnosed — and it is worse than a no-op: it **stops the music**

Picked this up while diagnosing #882 (which was written on this issue's "no-op" premise and reached a wrong conclusion because of it). The gesture is real and still live, but the symptom recorded here is not what happens.

**Ctrl+Enter while playing does not silently do nothing. It stops playback.**

### Root cause — it is not the `fireEvaluateSuccess()` path

This issue's hypothesis was that `onPlay` → `runtime.play()` doesn't fire `fireEvaluateSuccess()` on the second call. `play()` is **never called on the second press at all**. `StrudelEditorClient.tsx:1316` wires the action as a Play/Stop toggle:

```js
onPlay: () => {
  if (state.isPlaying) handleStop(tab.fileId);   // ← while playing: STOP
  else handlePlay(tab.fileId);
},
```

Instrumented, the second press logs `handleStop CALLED` and never `handlePlay`. `EditorView.tsx:369` registers it under the label `'Play / Stop'`, so the toggle reads as deliberate at the call site. Present since `d605ada7` (2026-04-10, phase 10.2 wiring) — a contract that drifted, not a regression.

[...]

PR #892, opened 2026-07-15
📝 docs(tests): Ctrl+Enter while playing stops — it is not a no-op (#180)

MrityunjayBhardwaj:
## Problem

Two spec comments record that `Ctrl+Enter` while playing is a **silent no-op**, citing #180:

- `strudel-viz-methods.spec.ts:97` — *"Manual Ctrl+Enter while playing is a no-op in this codebase"*
- `perf-matrix.spec.ts:207` — *"Ctrl+Enter WHILE PLAYING is a no-op (#180)"*

It is not a no-op. **It stops the transport** — the action is wired as a Play/Stop toggle (`StrudelEditorClient.tsx:1316`).

That distinction is not pedantic. #882 was written on the "no-op" premise and concluded that viz options were being dropped on re-evaluate. They are not: a gesture that silently *stops playback* looks exactly like an options bag arriving empty, because the sketch stops painting either way. With live mode on — the path that genuinely re-evaluates while playing — an edited option applies correctly (observed: red `0.655` → green `0.700`). #882 is closed as not-a-bug.

## Fix

Say what the gesture actually does, and why each spec's `stop+play` is load-bearing rather than incidental. Both workarounds were already correct and are unchanged — only the stated reason was wrong.

## Not in scope

The product behaviour is tracked in #180, which I've updated with the diagnosis and retitled (it also turns out to break the documented getting-started walkthrough at step 3). This PR only stops the false premise from being read as settled fact by the next person — it changes no behaviour and no assertions.

## Test plan

- [x] `strudel-viz-methods.spec.ts` — 26/26 green
- [x] Comment-only diff; no assertions, no product source touched

### DIFF
--- a/packages/app/tests/perf-matrix.spec.ts
+++ b/packages/app/tests/perf-matrix.spec.ts
@@ -204,8 +204,9 @@ async function runCode(page: Page): Promise<void> {
 }
 
 // Ctrl/Cmd+. = stop (EditorView.tsx:356). MUST stop before re-evaluating —
-// Ctrl+Enter WHILE PLAYING is a no-op (#180), so without a stop every other
-// rung's eval is swallowed and measures the previous scene.
+// Ctrl+Enter WHILE PLAYING stops instead of evaluating (it is a Play/Stop
+// toggle, StrudelEditorClient.tsx:1316, #180). Without a stop, presses just
+// alternate play/stop and every other rung measures the previous scene.
 async function stopCode(page: Page): Promise<void> {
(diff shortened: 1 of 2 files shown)


MrityunjayBhardwaj:
Superseded by #893, which folds these corrections in. Keeping them separate would have been worse than useless: this PR documents that "Ctrl+Enter while playing STOPS", and #893 changes exactly that — so the comments would have been accurate for about a day, and the two PRs touch the same lines. #893 states the post-fix truth instead.

Issue #880, opened 2026-07-13
Viz options are silently dropped in the worker renderer — every .viz(name, {…}) option has been dead on the default path


MrityunjayBhardwaj:
## Problem

Every per-render viz option is **silently ignored** whenever a sketch renders in the viz worker — which is the **default path** for all users since #245.

```js
$: note("c3 e3 g3 c4").s("sawtooth").viz("pianoroll", { background: '#cc1133' })
```

The background stays default. So do `labels`, `vertical`, `fold`, `minMidi`/`maxMidi`, and any option a user passes to their own sketch via `stave.options`. The sketch just quietly draws its defaults — no error, no warning, nothing in the console.

Affects all three option syntaxes:
- `.viz("name", { … })` (inline)
- `._pianoroll({ … })` (underscore inline)
- `.pianoroll({ … })` (backdrop)

## Root cause

The options bag never crosses the worker boundary.

- Main thread — `P5VizRenderer` hands `components.options` to the p5 compiler through a live ref (`P5VizRenderer.ts:125`, `:220`), so `stave.options` works.
- Worker — `WorkerVizRenderer` marshals the scheduler, the analysers, the aliases and the vizConfig subset… but **not** `components.options`. The mount message has no field for it, and `hostP5Worker.ts:352` builds the compiler's ref as a hardcoded empty literal:

```ts
const optionsRef = { current: {} as Record<string, unknown> }
```

So `stave.options` inside a worker-rendered sketch is **always `{}`**.

This landed the day worker rendering became the default (#245, 2026-06-04) and has been live ever since.

## Observed (same fixture, same assertion, only the renderer path differs)

| renderer path | `{ background: '#cc1133' }` → red fraction |
|---|---|
| worker (default) | **0.000** — option dropped |
| main thread (`stave.viz.worker=0`) | **0.754** — option applied |

## Why nobody noticed

The specs that guard this all read pixels with `canvas.getContext('2d')` — which **throws** `InvalidStateError` on a `transferControlToOffscreen` canvas. They have been failing since the same day (part of the 15-failure canvas cluster in #875), inside a red gate.

And the one spec that *claimed* to cover the plumbing — "the options object reaches the sketch" — only asserted that a canvas existed and nothing threw, so it stayed **green** for the whole six weeks. An assertion that cannot fail is not coverage.

## Fix

Thread the bag across the boundary the same way the aliases and vizConfig subset already are:
- `MountMessage.options` — send it at mount.
[...]

Issue #281, opened 2026-06-08
✨ feat: GLSL/ShaderToy renderer — STEP 2, validate the renderer contract (#279)


MrityunjayBhardwaj:
## Problem
The engine-agnostic `RendererStrategy` contract is now written (#279 PR #280,
`architecture/renderer-contract.mdx`). It CLAIMS the seven per-frame/per-mount
seams are kind-agnostic — but that's only proven by adding a new renderer against
it. Empirically building-then-documenting finds the contract by stumbling;
deductively writing-then-validating proves it (PK31).

## Solution — the simplest possible new instance
A single-pass **GLSL/ShaderToy** renderer: Tier-1, zero-library, no-DOM-shim,
pure WebGL2 (fullscreen triangle + fragment shader, renders DIRECT to the
transferred OffscreenCanvas — no blit). The simplest instance that still SPANS
the complexity space, so slotting it into the contract's seams validates the
abstraction.

**v1 scope (HARD — Vairagya):** single-pass `mainImage(out vec4 fragColor,
in vec2 fragCoord)` + `iResolution`/`iTime`/`iMouse`/`iChannel0` (the master
analyser uploaded as a 2-row audio texture: FFT + waveform). NO multipass /
buffers / cubemaps / keyboard / mic.

## The validation (the whole point)
Predicted: GLSL slots into all seven seams with NO new host seam — the audio
feed reads the already-refreshed `rawAnalyser` (worker) / `AnalyserNode` (main),
exactly as hydra does, and uploads to a texture internally. If confirmed →
contract proven. If a new seam is needed → catalogue it + AMEND the chapter.

## Plan
- `glslCore.ts` (shared WebGL2 core) + `glslShaderSource.ts` (ShaderToy wrapper)
- `mountGLSL` in hostP5Worker + `GLSLVizRenderer` (main, fallback path)
- `makeGLSLRenderer` + wiring: MountMessage.kind, vizCompiler, vizPreset,
  registry, defaultDescriptors, 1–2 builtin presets
- Observe: worker path paints + audio-reactive; fallback path; then catalogue
  the contract verdict + (STEP 3) the engine-specific GLSL dev chapter.

Part of #279. Follows PR #280.

MrityunjayBhardwaj:
Shipped to main via #321 (performance milestone merge). Landed on performance at 9a36c81 — GLSL/ShaderToy renderer STEP 2 (renderer contract validated).

PR #895, opened 2026-07-15
✨ feat(viz): give the options invariant one home, and let hydra read it (#883)

### LINKED ISSUE #883: Options passed to a hydra or GLSL viz are silently ignored (stave.options is p5-only)
## Problem

`stave.options` only exists for **p5** sketches. A user can pass an options bag to a hydra or GLSL viz and it is silently ignored — on the main thread and in the worker alike:

```js
$: s("bd*4").viz("myhydrasketch", { intensity: 0.8 })   // the sketch cannot read `intensity`
```

The p5 compiler wires an `optionsRef` into the sketch scope (`p5Compiler.ts`), so `stave.options` resolves. The hydra bag (`hydraStaveBag.ts`) and the GLSL uniform set have no equivalent — nothing carries the bag, and the sketch has no way to reach it.

## Not a regression

This is pre-existing and renderer-consistent (options were never wired for these kinds), unlike #880 where p5 options existed and were dropped only in the worker. Filing it because it is the same **silent** class: the option is accepted at the call site, nothing warns, and the sketch just draws its defaults.

## Options

1. Wire the bag into the hydra bag + GLSL uniforms (`stave.options` everywhere — consistent with p5).
2. Or reject/warn on options passed to a kind that cannot consume them, so it fails loudly instead of silently.

Low priority; raised during the self-review of #881.

MrityunjayBhardwaj:
Closes #883. GLSL's half is #894.

## Problem

"Expose `components.options` to the sketch" was a **rule, not a structure**. `VizRenderer` is an interface, so each of six renderers had to remember it independently — in **two** places each, since the bag arrives on `mount()` and is re-published on every `update()`. Twelve chances to forget.

Three forgot. And this exact fault already shipped: **#880 was the worker renderer omitting the same field**, killing every viz option on the default path for six weeks. Forgetting is invisible — a dropped option is pixel-for-pixel identical to a sketch drawing its default, so nothing errors and nothing warns.

One invariant spanning six classes, enforced nowhere, isn't three bugs. It's a missing home.

**The bag was never missing.** `EngineComponents.options` already documents that it *"flows engine → component bag → **renderer** → `stave.options`"*, and `HydraVizRenderer.mount(…, components, …)` has always **received** it. It was delivered to every renderer and dropped at the leaf. (My original framing on #883 — "nothing carries the bag" — was wrong; I'd cited my own issue text instead of re-deriving it from the code.)

## Fix

**`VizRendererBase`** seals `mount`/`update` and captures the bag before delegating to `onMount`/`onUpdate`. A renderer *cannot* silently drop options, because it never handles the lifecycle call that carries them — including the next renderer someone adds, which is the case review would never catch.

- The slot is a live ref whose `.current` is **replaced**, never merged — so a removed key stops applying (#881).
- P5 / hydra / GLSL / worker migrated. The decorators (`FallbackVizRenderer`, `TeardownOnPauseRenderer`) deliberately don't extend it: they forward `components` verbatim to an inner renderer whose own base captures.

**Hydra exposure** — `stave.options`, a getter reading *through* the slot. The bag is built once per mount and mutated in place, so a captured value would pin the first evaluate's options forever.

**The worker too** — `hostP5Worker.mountHydra` built an options-less bag and `setOptions` was p5-only. Wiring only the main thread would have reproduced #880 exactly: options working inline and dead full-screen. (`postOptions` was already kind-agnostic, so the bag was crossing; the host just ignored it.)

## Not in scope

GLSL (#894). A shader can't read a JS object, so options must become uniforms — and the v1 uniform set is closed on purpose (#281, "the user never declares uniforms"). That's a design task with real edges (name collisions, key-change recompiles, strings that can't cross), not a patch to ride along here.

## Test plan

- [x] **New e2e** drives a real hydra sketch through `.viz("name", { … })` on the **default worker path**, reading pixels via the compositor (a transferred canvas throws on `getContext` — #875). The probe *paints* its options into colour channels rather than using a `window.__x` channel, which doesn't exist off-main.
- [x] **Falsifiable both ways** — breaking the getter turns both e2e specs RED, painting black: the exact silent failure being fixed. Restored → green.
- [x] **Full gate: 540 = 409 passed / 131 skipped / 0 failed**, reconciled (407 + 2 new). This refactor touches every renderer's `mount`/`update`, so units alone were not verification.
- [x] editor **2955** (2944 + 11 new unit tests, incl. "a subclass that never touches options still gets them captured")
- [x] app **717**
- [x] editor `tsc`: **73 errors vs 76 on base**, none in any file touched here
- [x] `dist` rebuilt and staged whole — the `HydraStaveBag` type change renamed a DTS chunk, which a named-file `git add` would have stranded


MrityunjayBhardwaj:
## Self-review — one finding worth surfacing rather than burying

**`mountHydra`'s `setOptions` is NOT exercised by the new e2e.** I checked instead of assuming: neutering the line leaves both specs green. An inline re-eval **remounts** the zone, so the edited bag rides in on `MountMessage.options` and the control-message path is never reached.

That matters because it's the shape of #881's bug — a silent no-op "fix" for a silent no-op bug, which only a test that could actually fail would have caught. So, explicitly:

- **What the e2e proves:** a real hydra sketch reads `stave.options` from its `.viz()` call on the default worker path, and an edited value applies. Both go RED when the getter is broken (the sketch paints black — the true silent-failure signature).
- **What it does not prove:** the no-remount update path.

I kept the line rather than dropping it as unproven, because it is load-bearing against exactly the class this PR exists to kill: `WorkerVizRenderer.update` posts an `options` message for **every** kind, and the dispatcher does `state?.setOptions?.(msg.options)` — so without it, a hydra options message is silently swallowed. The second commit says all of this in the code, so the next reader doesn't inherit coverage that isn't there.

Worth noting the same likely applies to `mountP5`'s `setOptions` (#881): every existing option spec stops+plays or re-evals, and both remount. If someone wants that path genuinely covered, it needs a fixture that changes options **without** changing the sketch — which is its own small piece of work, not something to fake here.

### Other checks

- **No renderer missed.** The only remaining `implements VizRenderer` are the two decorators (`FallbackVizRenderer`, `TeardownOnPauseRenderer`), which correctly forward `components` to an inner renderer whose base captures — extending the base there would capture a bag they never expose.
- **`WorkerVizRenderer`** inherits the slot but still marshals from `components` in `postOptions`. Harmless redundancy, left alone: `postOptions` must stay pure (it burned #881 once already).
- **`GLSLVizRenderer`** now captures options and does nothing with them — deliberate. The slot is ready for #894; the base is what makes that a leaf change instead of another thread-it-through.

Issue #422, opened 2026-06-16
Song timeline: expand + bind — read-only inline notes → Pattern panel


MrityunjayBhardwaj:
Part of #416 (canvas clip-timeline milestone) · roadmap **item A** · canvas-plan **Phase 3**.

## Problem
The canvas Song timeline (shipped via #419) renders per-track lanes with coarse density + capped mini-note marks, but a lane is opaque: you can't see a track's notes in detail, and there's no way to jump from a lane into the existing Sequencer / Piano Roll to edit it. The arrangement view and the note editors are disconnected.

## Solution
Click/expand a lane (a "clip" = a track's pattern chunk for now) to:
1. **Read-only inline note detail** — accordion the lane taller and draw its marks at higher detail (taller marks + pitch spread + per-beat gridlines). Viewing only — no inline editing (PV117: one note editor, in the panel).
2. **Bind** that pattern into the existing Pattern panel (Sequencer / Piano Roll) — reuse the existing cursor→panel binding seam; do not build a second editor.

Multi-expand allowed (cross-track alignment).

## Scope / non-goals
- Display granularity stays **per-track lanes** (true clip segments via cat/arrange = a later phase).
- No in-place inline note **editing**; all edits route to the Pattern panel via the bind.
- Worker rendering is a separate, profiled-jank-gated phase.

## Approach
- **Hit-test** (clientX/Y → lane + content-cycle) built from the scene + shared transform; pure + unit-tested.
- **Expand state** (Set of expanded lanes); generalize the uniform row-height assumption to a per-lane layout (top/height) feeding draw, canvas host height, lane labels, and hit-test together.
- **Bind wire**: map a lane → a representative source location → move the editor cursor so the panel re-detects the active chunk.
- **Gesture**: keep seek-anywhere; add a lane disclosure caret + double-click to expand/bind (DAW convention: ruler/body = seek, caret/double-click = open detail).

App-only (no editor dist rebuild). Tests + screenshot observation.

MrityunjayBhardwaj:
Shipped via PR #423 (`feat/song-timeline-expand-bind`, merged to `studio_v0.2.0`); verified on current studio.

Both halves of the solution are in `FullSongTimeline.tsx`:

[...]

PR #896, opened 2026-07-16
✨ feat(timeline): let a note remember where it came from (#874 phase 1)

MrityunjayBhardwaj:
Phase 1 of the three-phase plan filed on #874. Pure data — **nothing renders or clicks differently yet.** Phases 2 (the hit-test) and 3 (the wiring + un-skipping the spec) follow separately; this is deliberately the boring one.

## The problem

Clicking a note on the Song timeline used to jump the editor to that note's source. It stopped working when the live view moved from DOM to canvas, and no commit removed it — `git log -S` finds nothing to find.

In the DOM view a drawn note **was** its `IREvent`, so `evt.loc[0].start` was always in hand and click-to-source came free from identity. The canvas draws a `SceneNote` — a projection that kept the geometry and dropped the link. Nobody forgot the constraint; it was free under the old representation, so it was never a constraint at all.

The offset was never actually missing: `timelineMarks` holds the event at **both** push sites, and the eval path already reads `ev.loc?.[0]?.start` twenty lines later to attribute the note to a lane. It was computed, used, and thrown away.

## The fix

Carry it. `SceneNote.sourceOffset` is set from the event's location at both push sites, and the eval path now reads that offset **once** for the two questions it answers — which lane owns this note, and where its source is — instead of computing the same expression twice.

**The field is required, not optional like `voice`.** That difference is the point. A missing `voice` means something harmless (group under `NO_VOICE`); a missing `sourceOffset` silently re-drops the exact capability this field exists to restore. Required means a future mark source can't omit it without the compiler objecting — a guarantee that holds for someone who never read any of this. Fixtures that don't care say `null`, which honestly states "no source". The compiler also confirmed the enumeration was complete: every one of the 45 resulting errors was in a test, so those two push sites really are the only places production builds a `SceneNote`.

## Observed, not assumed

Evaluating through the **real transpiler** (the engine's own call shape) rather than reading the code:

```
note("c3 e3 g3")  → c3@[6,8]="c3"  e3@[9,11]="e3"  g3@[12,14]="g3"
stack(\n note("c3 e3"),\n s("bd*2")\n) → c3@[15,17]  e3@[18,20]  bd@[29,31]
```

Each hap's `loc[0]` lands exactly on its own token, in document coordinates, correctly across lines. (A first probe called `@strudel/core`'s bare `evaluate` and got mini-string-relative offsets — `[1,3]` → `"ot"` — which would have "proved" the offsets were garbage. The transpiler does the shifting; the instrument was wrong, not the data.)

## A residual worth knowing before phase 2

That same observation turned up something the plan didn't anticipate, and it **falsifies the "`[0]` is the innermost token" premise** I had started to write into the code:

| pattern | `locations` in order | note's own token |
|---|---|---|
| `note("c3 e3 g3")` | `"c3"` | `[0]` ✓ |
| `s("bd*2")` | `"bd"`, `"2"` | `[0]` ✓ |
| `.add(12)` / `.fast(2)` / `.rev()` | `"c3"` | `[0]` ✓ |
| `.add("<12 7>")` | **`"12"`**, `"c3"` | `[LAST]` ✗ |
| `n("0 2 4").scale("C:major")` | **`"major"`**, `"C"`, `"0"` | `[LAST]` ✗ |

A transform whose argument is a **mini string** contributes its own locations and they land first, so the note's token goes last. For `n("0 2 4").scale("C:major")` — a canonical idiom — all three notes report the same offset, the scale argument. The degree token each note wants is in the array, at the end.

`loc[0]` is kept here because it's what the filed plan specifies, it's correct for that plan's three acceptance cases, and it matches what `sourceByLane` already ships. That last point cuts both ways: **a scaled track's expand-to-bind (#422) anchors on its scale argument today**, same root, pre-existing.

Phase 2 has to decide which element to read — that's where the hit-test makes the choice observable end-to-end. A test pins the current behaviour explicitly as a known residual so that decision starts from a fact instead of an assumption; **it is meant to flip** when phase 2 picks a rule.

## Test plan

7 new tests, **each shown red before green** — this repo has shipped greens that couldn't fail, so every one was falsified:

- **Neuter both push sites to `null`** → 4 (now 5) fail; zero pre-existing tests break, so the arm isolates the field alone.
- **Invent an offset (`0`) instead of reporting `null`** → the two null-residual tests fail. They can't be caught by the arm above, so they get their own.
- **Read the outermost loc instead of `[0]`** → exactly 1 fails.

The load-bearing one asserts each note gets its **own** offset, diverging from its lane's shared anchor — a per-lane value would collapse every note to one destination, which is just the lane-level jump we already have.

- App suite: **724 passed** (717 + 7, reconciled)
- `tsc` clean; editor package untouched, so no dist rebuild
- Fixture edits are mechanical `sourceOffset: null` additions at compiler-flagged lines

Part of #874 (phase 1 of 3) — does not close it.

### DIFF
--- a/packages/app/src/components/__tests__/drawTimeline.barscale.test.ts
+++ b/packages/app/src/components/__tests__/drawTimeline.barscale.test.ts
@@ -40,7 +40,7 @@ function scene(): TimelineScene {
   return {
     lanes: [{
       laneKey: 'd1', displayName: 'd1', color: '#7af', density: [1],
-      notes: [{ cycle: 0.1, end: 0.2, pitch: null, gain: 1 }],
+      notes: [{ cycle: 0.1, end: 0.2, pitch: null, gain: 1, sourceOffset: null }],
       pitchMin: null, pitchMax: null, voices: [],
       clips: [{ armIndex: -1, startCycle: 0, endCycle: 1, label: null }],
       sourceOffset: null, arrangeOffset: null, labelOffset: null,
--- a/packages/app/src/components/musicalTimeline/__tests__/collectNoteMarks.test.ts
+++ b/packages/app/src/components/musicalTimeline/__tests__/collectNoteMarks.test.ts
@@ -131,3 +131,142 @@ describe('collectNoteMarks — eval-backed marks (#861)', () => {
     expect(marks.labelOffsetByLane.get('d2')).toBe(30)
   })
 })
+
+/**
+ * `sourceOffset` — the note's link back to its own source token (#874 phase 1).
+ *
+ * A `SceneNote` is a PROJECTION of an `IREvent`. In the retired DOM live view a
+ * drawn note WAS its event, so `evt.loc[0].start` was always in hand and
+ * click-to-source came free from identity; the canvas projection kept geometry
+ * and dropped the link, which is how the gesture vanished with no commit that
+ * removed it. These tests hold the link that restores it.
+ *
+ * Phase 1 is DATA ONLY — nothing renders or clicks differently yet. What is
+ * asserted here is exactly what phase 2's hit-test will read.
+ */
+describe('collectNoteMarks — sourceOffset (#874)', () => {
+  it('carries each IR mark’s own source offset (pre-eval fallback)', () => {
+    // No eval events → IR path. d1's event is located at char 10, d2's at 40.
+    const marks = collectNoteMarks(null, { fake: true } as never, 4)
+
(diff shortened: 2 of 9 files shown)


MrityunjayBhardwaj:
Deferring per-note click-to-source restoration (#874) for now — closing this without merging. The work is intact on `feat/scene-note-source-offset` (commit 7d2322e7, pushed) and can be reopened as-is when we pick it back up. Nothing here is lost; this is a park, not a discard.

Issue #143, opened 2026-05-15
stripParserPrelude: guarded boot expression 'typeof X !== undefined && X(...)' not stripped (20-15 V-1 backlog)


MrityunjayBhardwaj:
## Problem (20-15 V-1 real-world backlog — NOT fixed this phase)

Sample `-7LU6zgzViSM` ("Doubly-Linked Liszt") opens with the defensive idiom:

```
typeof setDefaultVoicings !== 'undefined' && setDefaultVoicings('legacy') // ...
// @title ...
stack( ... )
```

The first line is a pure boot side-effect guarded by a `typeof` short-circuit. `stripParserPrelude` has no rule for a `typeof X !== 'undefined' && X(...)` guarded-call statement, so the program falls to Code-fallback even though the musical `stack(...)` below is fully parseable.

## Scope

Backlog — 1/14 V-1 fallbacks, low frequency but a recognisable idiom class. Per AnviDev: issue before fix. NOT a matcher→interpreter risk (it is a *recognition* extension of the boot-call line classifier, not evaluation).

MrityunjayBhardwaj:
Closed by PR #154 (merge aaae98c, 2026-05-18). GitHub auto-closes only one `closes #N` per keyword (#142); the rest are closed manually — the documented 20-15 lesson. Verified on main: editor 1564/1564, parity-corpus 32/32, loc-fidelity 32/32, real-world Bakery parity 72.0%→80.0% (N=50, sha f73b3956).

MrityunjayBhardwaj:
Superseded by #163 — 20-21 RESEARCH §R-1 9-wave bisect proves the actual blocker for `-7LU6zgzViSM` is chain-arg walker `//`-comment apostrophe handling (`findMatchingParen` + `splitArgsWithOffsets` test string delimiters BEFORE checking for line comments). The 20-16 `GUARDED_BOOT_RE` ship is correct for its idiom class but does not reach this exemplar's actual mechanism (the blocker is 3 calls down the parser pipeline, inside `.layer(arrow1, arrow2)`).

#143 stays closed; #163 tracks the correct class.

Issue #142, opened 2026-05-15
stripParserPrelude: samples({...}) / samples('github:...') multi-line + object-literal boot args not stripped (20-15 V-1 backlog)


MrityunjayBhardwaj:
## Problem (20-15 V-1 real-world backlog — NOT fixed this phase)

`samples` IS in the prelude skip-set, but the V-1 Bakery pull surfaced two boot-call shapes that still fall through:

1. `samples({ o0: 'samples/ocean/ocean_00.wav', o1: ... })` — multi-line **object-literal** arg (sample `-P398OK_eprf`).
2. `samples('github:yaxu/clean-breaks')` / `samples('https://raw.githubusercontent.com/...')` co-occurring with `var cpm=30; stack(...)` (samples `-1j62z5xjyCN`, `-6c1hEXe8Agi`, `-HyFCSbuSlq5`) — here the binding (#141) is the primary blocker, but the multi-line `samples(...)` prelude recognition is a secondary contributor.

The `PRELUDE_CALL_RE` line-classifier recognises single-line `samples(...)` calls; a brace-spanning object-literal arg isn't depth-walked as a single prelude statement (it currently relies on the multi-line boot-call depth walker only matching the enumerated call shapes).

## Scope

Backlog. 1–3 of 14 V-1 fallbacks. Per AnviDev: issue before fix; D-04 `${}`-interp etc. stay correct Code-fallbacks.

MrityunjayBhardwaj:
Closed by PR #154 (merge aaae98c, 2026-05-18). Resolution: **fixture-only** — B-1 OBSERVED (vite-node, verbatim issue body) that the existing stripParserPrelude depth-walker (pS:182-250) already brace-balances multi-line `samples({…})` + recognises `samples('github:…')`/`samples('https://…')` string forms. No code change needed; the `github:` residual Code-fallback is the co-occurring binding (#141, → Phase 20-17), NOT a #142 strip gap. `bakery-142-samples-objlit.strudel` vendored as the permanent regression wall (parity-corpus 32/32 green on main).

Issue #141, opened 2026-05-15
parseStrudel: var-keyword bindings + binding refs outside stack() args → Code-fallback (20-15 V-1 backlog)


MrityunjayBhardwaj:
## Problem (20-15 V-1 real-world measurement, D-03 backlog — NOT fixed this phase)

The V-1 fresh Bakery pull (N=50, 2026-05-15, upstream SHA `f73b3956`) measured **72.0% real-world structural parity** (36/50) vs the 4/10=40% baseline. Of the 14 Code-fallbacks, the **dominant class (6/14)** is binding shapes γ-3 intentionally did NOT cover:

- `var x = ...` (γ-3's regex is `let|const|var` for the *binding-map build* but the measured fallbacks reference the binding **outside** a `stack()` arg — e.g. `const beat = sound(rp1)...`, `note(rp1)`, `n(chordProgression)`, `const both = "...".s()` then `stack(drum,clap,bass,perc)` where an arg is itself a *binding whose RHS references another binding*).

Representative samples (gitignored artifact `samples-2026-05-15T23-13-07-584Z.json`): `--LsnlgQ6osk`, `-1j62z5xjyCN`, `-72eEl7NwK9e`, `-CyO42BOyp5a`, `-L13nBhrqGR_`, `-LHtBlF8peGC`.

## Relationship to #140

This is the real-world frequency evidence for #140 (the dropped γ-4 STRETCH: generalize binding-map substitution beyond `stack()` bare-ident args). #140 is the fix vehicle; this issue records that V-1 measured it as the **single highest-frequency remaining gap class (6/14 ≈ 43% of fallbacks)** and supplies the concrete repros. The D-02 matcher-not-interpreter boundary still applies (forward refs / reassignment / arrow RHS stay graceful Code).

## Scope

Backlog — friction-first next cycle. Per AnviDev: issue filed before any fix. Bumps the priority of #140 with measured frequency.

MrityunjayBhardwaj:
Closed by PR #157 (single closes-keyword GitHub limit — #140 took the auto-close slot; the 20-15/20-16 lesson). Verified merged via git merge-base ancestry: 98edc8ac201a68bc41b120e3df1f353c6ac08ccf is a true ancestor of origin/main; merge commit e3e6571. Artifact verified on real main: editor 1603/1603, app 367/367, parity 33/33, loc-fidelity 33/33, wave anchors present in built dist.

MrityunjayBhardwaj:
[...]

PR #898, opened 2026-07-16
🩹 fix(parity): make the fallback classifier tell the truth about cause (#897)

MrityunjayBhardwaj:
Fixes the instrument this session kept getting misled by. Closes #897.

## The problem

The parity sampler's cause classifier (`_bakery-classify.spec.ts`) ended with a last-resort arm:

```js
if (/\b(let|const|var)\s+…=/.test(live)) return '…binding ref outside stack()-bare-arg'
```

It returned "binding ref" for **any** Code-fallback whose source merely *contained* a `let`/`const`/`var`. That made binding-ref look like the dominant real-world blocker (~59% of fallbacks) and pointed the roadmap at a fix that does nothing.

Two things make it a phantom:
- **`parseStrudel` already resolves bindings** — `const`/`let`/`var`, transitive 3-hop chains, method-ref chains, and refs in every position (receiver, numeric arg, whole-expr). Verified directly against the parser.
- **Reading the 20 patterns it bucketed as "binding"**, the binding was incidental in every one. They fall back for Hydra visual code (`initHydra`/`.out()`), floatbeat DSP (`` dough`…` ``/`Function(…)`), arrow-fn/lambda, `function`/`register` definitions, or an unmodelled combinator (`stepcat`).

## The fix

Extract `classifyFallback` into a hermetic module and attribute the **real** cause:

1. **Correct-by-nature fallbacks first** — Hydra, DSP, lambda, function/class/register. These have no musical timeline by nature; no parser work recovers them, so they are not gaps.
2. **Then the addressable gaps** — `#142` samples-objlit, `#143` guarded-boot, module syntax, paren-root.
3. **No "binding ref" bucket.** A fallback that still carries a binding is a *structural residual* whose label states plainly that the bindings resolve and the real cause is other structure.

Ordering is correct-by-nature-first on purpose: a gap only counts if the named fix would *actually* recover the pattern — which keeps the GAP tally from over-counting the way the binding arm did (the inverse of the old error).

## Verification

- **`classifyFallback.test.ts`** pins the honest labels against real observed shapes (the fixture hashes are the actual samples). Shown **RED against the old logic (7/8 fail) → GREEN (8/8)**. Hermetic (no parser import), so it runs in the CI gate — the classifier's honesty is now a permanent regression test.
- Parity CI suite **123 passed**; `tsc` clean.
- **Live N=360 sweep, before → after:**

| before (phantom) | after (honest) |
|---|---|
| #141 binding: **20/34 (#1)** | binding: **0** (bucket removed) |
| — | functional/lambda 12, Hydra 8, DSP 2 → **22/34 correct-by-nature** |
| — | 6 triage + 4 structural residual + 2 samples-objlit → **12/34 addressable** |

The headline the backlog now tells: ~65% of fallbacks are correct-by-nature (no timeline exists), and the addressable remainder is dominated by individual triage, **not** binding resolution. The metric no longer points at a wall that isn't there.

### DIFF
--- a/packages/app/tests/parity-corpus/_bakery-classify.spec.ts
+++ b/packages/app/tests/parity-corpus/_bakery-classify.spec.ts
@@ -17,6 +17,7 @@
 import { describe, it } from 'vitest'
 import fs from 'node:fs'
 import { parseStrudel } from '../../../editor/src/ir/parseStrudel'
+import { classifyFallback } from './classifyFallback'
 
 const SAMPLES = process.env.BAKERY_SAMPLES
 const RESULT = process.env.BAKERY_RESULT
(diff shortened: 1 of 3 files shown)


MrityunjayBhardwaj:
Heads-up from #901 (PR #902) — it lands right on this PR's histogram, and I think one arm here wants a second look before merge.

I ran **this PR's `classifyFallback`** against the same 360 pinned samples (offsets 0/250/500), with and without #901's parser fix. Identical rows, only the parser differs:

| class | this PR | this PR + #901 |
|---|---|---|
| correct-by-nature (Hydra 8 · lambda 12 · DSP 2) | 22 | 22 |
| **GAP #142: samples({…}) object-literal boot arg** | **2** | **0** |
| GAP: structural residual (bindings RESOLVE — real cause is other) | 4 | 3 |
| GAP: uncategorised — needs triage | 6 | 5 |
| **addressable total** | **12** | **8** |

parity 326/360 (90.6%) → 330/360 (91.7%).

Two things worth noting:

**The good news first — this PR's honest labelling worked exactly as designed.** One of the four recovered patterns (`-hOLDE1gYQmP`) landed in your `structural residual (bindings RESOLVE — real cause is other)` bucket. That label was *right*: the bindings did resolve, and the real cause was `stack (`. The old catch-all would have called it a binding gap and buried it. This bucket pointed straight at a real bug.

**The thing to look at — the `#142` arm looks like the same shape this PR was written to remove.** At `classifyFallback.ts:101`:

```js
if (/\bsamples\s*\(\s*\{/.test(live)) return 'GAP #142: samples({...}) object-literal boot arg'
```

That's a grep for a *feature of the input* (`contains samples({`), not a verified cause — the same predicate shape as the binding arm this PR deletes. Three things line up against it:

1. The only 2 patterns it labels (`36WbGIEN0uMZ`, `3MLO1QCnluxT`) are **`stack\n (`** programs. They also happen to call `samples({…})`, and that check runs before anything stack-related, so it absorbed them. After #901 the bucket is **empty**.
2. **#142 is CLOSED**, and its own corpus fixture (`bakery-142-samples-objlit.strudel`) parses structurally today — so the shape it names already works.
3. Same for the `#143` arm: it also names a closed issue, and labels **0** patterns in this sweep (your correct-by-nature ordering absorbs those three, correctly).

So after #901, both remaining named-issue GAP arms name closed issues and match zero real-world patterns. Options as I see them: drop the `#142`/`#143` arms and let those patterns fall to `uncategorised — needs triage` (which is honest — the sweep says there's nothing there), or keep them but re-verify against a live repro first.

Not touching this PR either way — #902 is independent and doesn't depend on the outcome. Just didn't want the `2 samples-objlit` line to carry forward into the roadmap when it's really `stack (`.

PR #906, opened 2026-07-17
♻️ refactor(notation): ask Strudel what the mini grammar is, don't re-derive it (#903)

MrityunjayBhardwaj:
Part of #903. **Does not close it** — this is A1 (read); A2 (span-surgery write) and A3 (widen the view) remain.

## The problem

`visualEdit/notation/parse.ts` hand-rolled Strudel's mini-notation grammar — `tokenize`/`parseGroup`/`readElongation`/`readMultiplier`/`readReplicate`/`readEuclid`, plus a copy of `bjorklund`. It re-implemented a grammar `@strudel/mini` already ships, complete and located.

Every real-world "gap" it reported turned out to be **drift from the original, never a missing feature**: krill parses **623 of the 625** units our copy rejected. (The 2 residuals are a truncated source and a floatbeat DSP expression that isn't mini-notation.)

The drift *regenerated per position*, because the copy made position-specific — one reader per place a token could appear — what Strudel makes uniform (`weight`/`reps`/`ops` are fields on **every** element). That's why every gap read *"works on an atom, fails on a group"*. And #905 had just shipped krill's own answer transcribed back as a regex: a third place that knows the rule, free to drift the moment upstream moves.

## The fix

**krill decides SYNTAX; the view decides what it can SHOW.** `tokenize` maps krill's AST onto `Step[]`, and 224 code lines of grammar go — including the `ATOM` regex #905 added and the `/[<>{}/?%.|]/` char-class that read *content* as *syntax* (an `_` in a name, a `.` in a decimal, a `-` in `blue-velvet`).

Euclid distribution now calls `@strudel/core`'s own `bjorklund`. Our copy agreed with it on all 152 (k,n) pairs to n=16 — which is exactly the point: a copy is correct the day it's written and free to drift after, and nothing would have told us.

Refusals are now the **view's own**, and name the feature they mean.

> **Reviewer note — the file GREW, 702→790 (224 code lines out, ~257 in).** The plan predicted −195; that estimate came from reading. The file didn't shrink, it got honest: a one-line char-class rejecting eight distinct features is short *because* it lied. Naming each cause costs lines, and the naming is the deliverable — see the histogram below. Worth a second opinion.

## Results — swept A/B over 360 **identical** pinned bakery tunes (3×120 @ offsets 0/250/500)

| | before | after |
|---|---|---|
| units editable | 1096/2226 (49.2%) | **1117/2226 (50.2%)** |
| tunes with an editable surface | 256/360 (71.1%) | **259/360 (71.9%)** |
| per-tune | — | **+21 gained / −0 lost** |

**The percentage was not enough, and nearly shipped a regression.** The first swept run read `+32` — a clean win. Per-tune diffing showed it was `+33 gained, −1 LOST`: `[c4,e4,g4,c5]*2` (a chord with a multiplier) stopped opening, netted away into a plausible number.

So the real instrument is a **differential at the `tokenize` seam** over the **1,501 distinct mini strings** in that corpus — old parser vs new, every divergence localized: **28 recovered / 0 regressed.**

That differential also caught **two silent-data-loss bugs in the adapter itself**: `rd:<1 3 2>` and `[[0,1]*<1!3 2> …]` were being **truncated** to `rd` / a bare chord — they'd have rendered as editable and then written the user's notation away on first edit. All three now refuse instead. *A refusal is recoverable; silent loss is not.*

### The catch-all dissolved (the actual deliverable)

`uses mini-notation features beyond the editable subset` — **392 units (185 step + 207 roll) → 0**, replaced by named causes. Most of that is honest **re-attribution**, not recovery: the old reasons *lied about cause*. `unsupported token "!"` blamed `!` for 80 units whose real cause was `!` on a **group**.

Two findings for #903's backlog, both verified by reading raw inputs, not the histogram:
- **`/n` slow-stretch ≈ 85 units** — a class that was *invisible* inside the catch-all, and **larger than the backlog's stated #1**. Confirmed by reading 123 raw occurrences (`drm/9`, `<sine>/64`, `lp:6/4`). Multi-cycle, so it shares A3's `<…>` machinery.
- `ir/parseMini.ts` still holds a **third** `bjorklund` copy and is the other hand-rolled parser at this boundary. Deliberately out of A1's scope — worth its own issue.

## The gate that was missing

`parity.test.ts` gates the **timeline** parser. **Nothing gated the edit path** — the only instrument that measured editability lived in a maintainer-only `.spec.ts` that never runs in CI. So A1 rewrote a parser with no corpus-level net.

- **`editCoverage.ts`** — the oracle extracted to a hermetic module, imported by both consumers, never re-derived (a second copy would answer confidently and diverge silently — the failure this corpus exists to catch).
- **`edit-parity.test.ts`** — in CI. Per fixture: how many units open, and the exact reason for every refusal. Catches a regression **and** an undecided widening, as a reviewable diff. Same drift policy as `parity.test.ts`: **never `-u` to make it green.**

**It didn't work at first.** Armed with the real `[chord]*2` regression, the new gate passed 3/3 — the 55 vendored fixtures contained no such shape. Green, in CI, hermetic, and **inert for the exact bug it was built for**. Hence `bakery-903-chord-multiplier` + `bakery-903-patterned-variant`, from real Bakery tunes. With those in, the arm goes red (`editable: 2 → 0`). Both directions are pinned — the truncation shape is snapshotted as *refused with its reason*, so making it truncate silently flips it to `editable: 2` and fails.

## Guarding the internals we now depend on (`1bc36916`)

The adapter reads krill's **private** AST (no export map, no types). An upstream change would be **silent** — read `undefined` for a weight, decide "beyond the editable subset", grid quietly stops opening tunes. 15 tests pin every fact the adapter relies on.

**Deliberately NOT pinning the version**, though that was the first instinct: every `@strudel/*` package pins `core`/`mini` **exactly** (1.2.6) among themselves, so the set upgrades atomically and `^1.0.0` tracks it as one instance. An exact pin our side would let `@strudel/transpiler` want `mini@1.3.0` while we hold 1.2.6 — **two copies of mini**, which registers `mini()` on `Pattern` by side effect. Two `Pattern`s is worse than the drift being guarded. Track the set, assert the shape.

## Test plan

Everything below was **run**, not inferred.

- **Playwright gate: 544 = 413 passed / 131 skipped / 0 failed** (reconciled). An earlier run showed 1 red in `viz-shared-pump-observe` — attributed by measurement, not assumption: **10 reads/frame under load vs 5.00 clean**, same branch and dist. I caused it by running `edit:coverage` concurrently.
- editor **2983** (+15 contract, reconciled) · app **736** (+7: 3 new + 2×2 from the new fixtures feeding the existing per-fixture gates) · tsc **73** (pre-existing baseline, unchanged)
- **Every new spec shown RED-then-green via an arm that isolates it** — including one arm that proved a contract assertion was *too strict*: `rotation: null → undefined` can't break us (the adapter reads `== null`), so the contract must not cry wolf. Narrowed to `?? null`. A gate that red-lights harmlessly is how the `-u` reflex gets trained.
- `EDIT-COVERAGE.{md,json}` regenerated (55→57 tunes).
- krill/euclid stay **external** in the built dist (tsup externalizes dependencies) — no bundle growth; 0 peggy bytes inlined. Dist rebuilt and committed whole.

### Snapshot drift call-out (per `parity.test.ts`'s policy)

The `parity`/`loc-fidelity` snapshot diffs are **411 pure additions, zero deletions** — the two new fixtures only. **No existing fixture's IR moved**, independently confirming this change stays inside the notation path.

## Review asks

1. The **line-count growth** above — is "the file got honest" the right read?
2. The **view/grammar line**: grammar misreadings (`gm_agogo`, `bd!`, `@0.5`) were fixed; view limits (nested groups, `!` on a group, alternation, polymeter, `?`) are **kept**, now stated honestly. A1 should not do A3's widening — did I hold that line?
3. `bd!` now parses as `!2`. The old tokenizer required digits and refused it; **observed** through real Strudel, `mini('bd!')` ≡ `mini('bd bd')` ≡ `mini('bd!2')` (same two haps). I changed an existing assertion that encoded the old refusal — flagging it explicitly rather than burying it.


MrityunjayBhardwaj:
Filed #907 out of this work — and it turned out to be more than the duplication smell noted in the description.

`ir/parseMini.ts` holds a **third** hand-rolled `bjorklund`, and it **disagrees with Strudel on 44/152 (k,n) pairs (29%)**. It is live on the *display* path (`parseStrudel.ts:19` → `parseMini.ts:331`), so the timeline draws a rhythm the audio does not play: `bd(5,8)` plays `10110110`, the IR draws `10101011`. Observed end-to-end against `queryArc`, not inferred.

It survived because **`bd(3,8)` is correct** — the canonical example everyone tests with. The drift starts at `k=4, n=6`.

Out of scope for this PR (different path, and #906 is large enough), but it is the concrete cost of the duplication this PR is about: three copies of one function existed, and the one nobody A/B'd against the original was wrong. The fix in #907 is the same one-import move this PR already made for the edit path.

PR #908, opened 2026-07-17
🐛 fix(ir): draw the euclid rhythm Strudel actually plays (#907)

MrityunjayBhardwaj:
Closes #907. **Stacked on #906** — base is `refactor/notation-krill-adapter`, not `studio_v0.2.0`. The dependency is real, not stylistic: this fix needs the `declare module '@strudel/core/euclid.mjs'` ambient that ships in #906 (`git show studio:packages/editor/src/strudel.d.ts | grep euclid` → 0 hits). **Merge #906 first**, then this retargets to studio automatically.

## The bug

`ir/parseMini.ts` hand-rolled its own `bjorklund`, and it computed a **different rhythm than Strudel** — 44 of 152 (k,n) pairs to n=16 disagreed (29%). The euclid **rotation was backwards** on top of that: Strudel's `_euclidRot` does `rotate(b, -rotation)` (a RIGHT rotation, `euclid.mjs:130-134`); we left-rotated.

Neither is dead code — `parseStrudel.ts:19` imports parseMini, `ir/index.ts:38` re-exports it, and the euclid expansion at `parseMini.ts:331` feeds the IR that draws the timeline. **The timeline has been drawing a rhythm the audio doesn't play.**

| pattern | Strudel plays | the IR drew |
|---|---|---|
| `bd(3,8)` | `10010010` | `10010010` ✓ |
| `bd(5,8)` (cinquillo) | `10110110` | **`10101011`** |
| `bd(4,6)` | `101101` | **`101011`** |
| `bd(3,8,1)` | `01001001` | **`00100101`** |

**Live on real tunes.** `flatrave`, `meltingsubmarine` and `sampleDrums` in our own parity corpus use `sd:1(3,4,3)`, `hh27(3,4,1)`, `bd(3,4,3)`, `sn(3,4,2)` — all rotated euclids, all drawn wrong until now.

## Why it survived three months

**`bd(3,8)` is one of the pairs the copy got right**, and `rot=0` is a no-op in either direction. The canonical example works. The docs example works. Every hand-check passes. The drift only begins at `k=4, n=6`.

And the test that should have caught it **had the right answer in its title**:

```js
it('5 over 8 = [1 0 1 1 0 1 1 0]', () => {
  // The exact distribution depends on the Bjorklund variant used —
  // we only assert the count + length.
  expect(r.length).toBe(8)
  expect(r.filter(Boolean)).toHaveLength(5)
})
```

`[1 0 1 1 0 1 1 0]` **is** Strudel's answer. The body asserted *"5 onsets somewhere in 8 slots"* — which the broken distribution satisfies. Someone knew the correct rhythm, found the code disagreed, and concluded *"there are multiple valid Bjorklund variants"* rather than *"our code is wrong"* — then weakened the test and left the truth in the title, where it sat being right while the body held no constraint.

There is no variant freedom here: **the grid is a view of what Strudel triggers**, so exactly one distribution is correct, and it's upstream's.

## The fix

Import the authority — `bjorklund` from `@strudel/core/euclid.mjs`, the same function `.euclid()` runs, so what's drawn cannot disagree with what's played *by construction*. 40 lines of transcribed algorithm go.

- **Degenerate ends stay ours** — upstream computes `steps - hits` as the zero-run length and throws when `hits >= steps`; callers here expect a `steps`-long mask.
- **The rotation sign is fixed at the call site.** `rotate` itself is a correct generic left-rotate with exactly one caller; only that caller's sign was wrong. Changing `rotate` would have been the wrong repair.

## Test plan

**The oracle is Strudel's own evaluator queried for real haps — never a table of rhythms typed from belief.** That distinction is the point: a hand-written table is exactly what failed here, because it would encode the same wrong belief. Only asking Strudel what it actually triggers can disagree with us.

`euclidAuthority.test.ts` compares `parseMini`'s `Play`/`Sleep` slots against `mini(pat).queryArc(0,1)` for 12 patterns — deliberately including `bd(3,8)` and `rot=0`, the cases that always worked, so a regression that breaks only the others still fails — plus the full (k,n) sweep to n=16 asserting **total** agreement.

**Shown RED via two independent arms** (these are two separate bugs):
- restore the wrong distribution → **7 fail**
- restore the wrong rotation sign → **4 fail**, and `bd(3,8,0)` correctly still passes (`rot=0` is a genuine no-op)

The `5 over 8` assertion is repaired to check its own title, and armed: feeding it the old distribution turns it red.

- editor **2997** (+14, reconciled) · app **736** · tsc **73** (pre-existing baseline)
- **Playwright gate: RUNNING at the time of writing** — I will post the reconciled count as a comment when it lands rather than claim a number I have not seen. Unit + snapshot evidence above is complete and was run.

### Parity snapshot moved — 3 tunes, and the diff IS the news

Per `parity.test.ts`'s drift policy, this is called out rather than `-u`'d past. **It moved to the truth**, verified two independent ways:

1. All five real-tune patterns above now match `queryArc` **exactly**, checked individually.
2. **Play count 397 → 397, Sleep 53 → 53** across the whole snapshot — onsets **moved**, none added or removed. That's the structural signature of a rhythm correction rather than a parse change, and it's the check that would have caught me if the fix were wrong.

## Notes for review

- **No e2e drives a euclid onto the timeline at all**, and the one euclid e2e that exists (`sequencer.spec.ts:212`) uses `bd(3,8)` — on the *sequencer* path, which was already correct. The bug had nowhere to be caught. I did not add a canvas e2e here (onset positions off a canvas need compositor capture, and the unit test compares the two representations directly — the comparison that was never made). **The e2e gap is real and worth its own issue.**
- Both `bjorklund` exports (`notation/parse.ts`, `ir/parseMini.ts`) are now thin wrappers over the authority. The dangerous part — the algorithm — is gone from both; what remains duplicated is a 2-line degenerate-end guard with a bounded blast radius. **Not consolidating that here** (different layers, and a shared module for one 4-line function is the premature abstraction). Flagging it as a judgement call rather than doing it silently.


MrityunjayBhardwaj:
**The red Vercel here is a false negative — filed as #909.**

Not this PR's code. The discriminator is one line in the build log, and it holds 4/4 with controls both ways:

| commit | `@stave/app:build` | duration | status |
|---|---|---|---|
| `c98e6092` (this PR, rebased) | `cache hit, replaying logs` | 15s | ● Error |
| `247a5a83` (this PR, **identical content**, pre-rebase) | `cache miss, executing` | 2m | ● Ready |
| `3f198ec` (studio, 2d ago — predates my work) | `cache hit, replaying logs` | 11s | ● Error |
| 12h-ago studio deploy | `cache miss, executing` | 1m8s | ● Ready |

Turbo replays the recorded log on a cache hit instead of executing, so Vercel's `onBuildComplete` is **printed but never invoked** → no output collected → Error, after printing a complete success log. The rebase made this branch hash to a warm cache; that is the *only* thing that changed between Ready and Error here.

Locally on `c98e6092`: `next build` → ✓ Compiled, Route (app) emitted. Playwright gate reconciles 544 = 413 passed / 131 skipped / 0 failed. Editor 2983, app 736.

Holding this PR rather than merging over a red check — the call belongs to the maintainer, and #909 explains why the check cannot currently be trusted in either direction.

MrityunjayBhardwaj:
**This PR's red Vercel is the false red from #909 — it is not broken.** Evidence from its own failing deploy (`dpl_5NJ2jt4sTVjvyXQpt1iexKfve1eN`):

```
@stave/editor:build: cache hit, replaying logs 4f0bfaa30af31da7
@stave/app:build:    cache hit, replaying logs dbd85a266c7bb046
@stave/app:build:      Running onBuildComplete from Vercel
Time:    1.146s >>> FULL TURBO
Build Completed in /vercel/output [8s]
status ● Error
```

Cache hit → log replayed → `onBuildComplete` printed but never invoked → no output collected → Error. **There is no error message anywhere in the log**, and `1.146s` is not a build; it is a replay. This is #909's exact signature, with none of the code in this PR involved.

#910 fixes the cause. Once it lands on studio, this branch needs to pick it up (rebase or merge studio in) before its deploy can go green — Vercel builds the branch head, so the `turbo.json` override has to be present on it.

Flagging rather than acting: I have not rebased this branch. Note also that its base auto-retargeted from `refactor/notation-krill-adapter` to `studio_v0.2.0` when #906 merged, which is correct.

PR #934, opened 2026-07-18
✨ feat(ir): classify chain controls from Strudel's registry, not a whitelist (Tier 2)

MrityunjayBhardwaj:
## Problem
Parsing used a ~24-entry allow-list of method names in `parseStrudel`'s method switch; anything off it fell through to an opaque `Code` wrapper. ~250 genuine Strudel controls — `clip`, `sustain`, `attack`, `release`, `lpenv`, `rlp`, and more — were off the list, so they opaqued, starving the mixer and the structural timeline/roll views of the controls a track actually applies. This is invariant 2 of the code↔view ideal state ("classification is a deny-list").

## Fix
Before opaquing an unrecognised method, ask `@strudel/core`'s own control registry — `isControlName`, queried **live** so it can never drift from the running version (no transcribed copy of the list) — whether the method is a control. If so, classify it as a `Param` through the existing universal arg parser, preserving the **user-typed token** so round-trip stays byte-identical (`.rlp(0.5)` re-emits `.rlp(0.5)`, never the canonical `.roomlp`).

Design choice: **additive** fallback. The 24 curated `FX`/`Param` arms — which carry per-control routing and arg gating we already trust — are untouched; only the `default:` opaque arm gains the registry lookup. This means `scale` (a `@strudel/tonal` fn) and `reverb` (a webaudio effect, not a core control), both handled by their curated arms, cannot regress to opaque. The literal deletion of the switch's method list is deferred to the adapter-consolidation phase (#929), which owns that refactor.

Anything the parser can't model — a lambda, a signal, a nested expression — still opaques, so the "wrap, never silently drop" residual is preserved. Genuinely unknown methods still opaque.

## Test plan
- New `parseStrudel.denyList.test.ts`: off-list controls (`clip`/`sustain`/`attack`) → `Param`; aliased control keeps the user token + round-trips; pattern-arg control classifies; `scale`/`reverb` still classify (regression guards); unknown method + lambda-arg still opaque.
- Editor suite: **3046 pass**. App suite: **763 pass**.
- Parity + loc-fidelity snapshots: **137 nodes/corpus reclassify `Code`→`Param`**, verified one-way (Code only drops, Param only rises, FX unchanged — zero `Param`/`FX`→`Code`).
- Tests that used a real control (`.release`) as their "unrecognised method" example were rewritten to a genuinely-unknown method so they keep exercising the opaque path — the premise the feature falsified, not the intent.
- **Observed** the real `parseStrudel → collect` pipeline: `s("bd*2").clip(0.5).sustain(0.3).attack(0.01).lpenv(4)` now surfaces `clip`/`sustain`/`attack`/`lpenv` on `evt.params` (the field the timeline/mixer read), instead of dropping them.

Part of #928. (`Closes` won't fire from a `studio_v0.2.0` base — I'll close by hand on merge.)


MrityunjayBhardwaj:
## Self-review pass

Two things came out of auditing the diff rather than trusting it green.

**1. Fixed here (`d040d2f2`) — an alias classified into a field nobody reads.** The classifier keyed each control by the token the user typed. `collect` maps a param key onto its event field (`params.s` → `evt.s`), so an alias keyed by its own spelling went nowhere: `.sound("bd sd")` produced a Param under key `"sound"`, leaving `evt.s` null and every consumer of it blind — a control that looked modelled in the view and read as absent downstream. Its mini also parsed as *note* names rather than sample names (wrong field, wrong per-event duration).

Now keyed by the canonical name (`sound`→`s`, `rlp`→`roomlp`), with the sample-key flag derived from that canonical name. Round-trip is untouched — `toStrudel` emits `userMethod ?? key`, so `.sound("bd sd")` still re-emits `.sound("bd sd")`, never `.s(...)`. Observed: `evt.s = ["bd"]` where it was `null`. Snapshot direction re-verified against the pre-change base: **Code −137 / Param +137 / FX ±0**, unchanged; 15 nodes now carry `userMethod: "sound"` with zero `"sound"` keys.

**2. Filed, not fixed — #935.** The 13 curated FX arms parse args with `parseFloat`, so `.room("0.3 0.5")` opaques while the off-list `.roomsize("0.3 0.5")` classifies: the more common control is now the more limited one. Not a regression (the FX arms always did this) — this PR just makes it visible. Fixing it means deciding whether a pattern-arg FX node stays `FX` or becomes `Param`, which belongs with the adapter consolidation that already owns collapsing these arms.

**On metrics, honestly:** `edit:coverage` is **unchanged** (52.6% tunes / 51% units) and should be — it measures notation-grid editability, a different axis from IR classification. The measurable result here is the parity movement (137 nodes/corpus `Code`→`Param`) plus controls now reaching `evt.params`, not a coverage lift. I'd rather state that than claim a number this change doesn't move.

Final: editor **3048 pass**, app **763 pass**, typecheck clean in the touched files.

PR #946, opened 2026-07-19
♻️ Phase 3a.1: parse mini-notation with krill instead of our own grammar (#943)

MrityunjayBhardwaj:
Replaces the hand-rolled IR mini-grammar (`ir/parseMini.ts`) with a lowering adapter over krill's AST — the same grammar authority the notation layer already uses. Closes #943, part of #929 / #925.

## Why

The IR re-implemented a language Strudel ships complete and located: a hand tokenizer plus byte-position operator scanning (`*n`, `@n`, euclid, `:N`). Every gap in it was **drift from the original, never a missing feature** — and the drift was silent. Nothing threw; each bug drew a plausible timeline the audio never played.

## What was actually wrong (found by running the corpus, not by reading)

| mini | drew (before) | plays (Strudel) |
|---|---|---|
| `bd!3` | `bd` + a phantom note **"3"** | three `bd` |
| `.5` | note **"5"** — `.speed(".5")` became speed **5** | `0.5` |
| `bd - sd` | a note literally named **"-"** | silence |
| `c4/2` | `c4` + a phantom note **"2"** | `c4` slowed |
| `bd*2, ~ sd` | one flat **sequence** | a **stack** (voices together) |
| `C:major` | `C` + a phantom note **"major"** | `C` |

The `,` one is the most instructive: the hand parser split commas only *inside* brackets, so a **chord played as an arpeggio**. Every note was present and only the *onsets* were wrong — which is exactly why no existing test caught it.

## What changed

- **Grammar → krill.** The tree is built from `@strudel/mini`'s AST, lowering its uniform ops model (`weight`/`reps`/`ops`) into the existing tags: euclid still expands eagerly, `*`/`/` → Fast/Slow, `?` → Choice, `@n` → Elongate, `!n` → sibling steps, `:n` → slice param. Tokenizer and byte-position operator handling deleted — **397 lines, down from 512**.
- **Alignment read uniformly.** One converter handles the top level and every bracketed group, so `a,b` means a stack wherever it appears. This is what the position-specific parser structurally could not do.
- **Spans derived, not copied.** krill's spans *tile* the input and the padding lands on **either side** depending on syntax (`bd sd` trails, `a@2 b@2` leads), so each token's range comes from its own start + length rather than krill's tiling end. `loc-fidelity` pins this.
- **Euclid helpers unified** in `ir/euclid.ts`. The two copies had already drifted — notation inverted a negative pulse count, the IR returned empty — so one home also settles which behaviour is Strudel's.

## Verification — the oracle is Strudel, not us

New `krillAdapterHaps.test.ts` asserts the IR carries exactly the notes `@strudel/mini` *triggers*, for every corrected class. This matters: a test asserting our own output would have passed against the broken parser forever. Same discipline as the euclid-authority gate.

**32 corpus snapshots move; all 16 fixtures triaged, every move a correction** — `Seq`→`Stack` (comma), `Seq`→`Cycle` (`|`), restored decimals, `!n` replication, `-`→`Sleep`, `/n`→`Slow`, absorbed `:word` tails. No move was left unexplained; two are *narrowings* that remove phantom notes without yet rendering the feature (patterned `:` tails like `rd:<1 3 2>`, and euclid on a group `[4 6](3,4,2)`) — both previously drew the operator's **arguments as notes**.

krill also rejects malformed mini the hand parser invented output for (`bd(3)`, `{}`); those degrade to the opaque fallback instead of a fabricated note, since Strudel itself refuses them.

## Gates

- editor **3069/3069** · app **763/763** · parity-corpus **161/161** · `gate:editing` both halves green
- notation gates (`mini-corpus`, `round-trip`, `edit-locality`, `writer-reach`, projection stress) **untouched and green** — the change stayed confined to the IR
- tsc unchanged at **73 pre-existing** errors; every touched file is clean
- editor `dist` rebuilt and committed

## Follow-ups worth filing

1. Patterned `:` variants (`rd:<1 3 2>`) — index alternation no longer displayed.
2. Euclid on a group (`[4 6](3,4,2)`) — rhythm not applied.
3. Weighted alternation arms emit a per-cycle retrigger where Strudel emits one hap spanning the arm. Pre-existing in `collect`, pinned by `weightedSlowcat.test.ts` — belongs to the engine phase (#945).


MrityunjayBhardwaj:
## Observation pass — drove the real app, before vs after

This PR was verified by tests (editor 3069 · app 763 · parity-corpus 161, plus a hap oracle against Strudel's own evaluator) but never actually *observed*. Closing that gap: I drove the running app on both this build and its parent (`9d62890a`), whose committed `dist/` gives a genuine pre-change baseline.

### No regression — and now it's observed, not inferred

Boot → seed → evaluate → expand lanes → screenshot, for all five fixtures the change moved (`sampleDrums`, `holyflute`, `dinofunk`, `flatrave`, `arpoon`).

Lane and voice structure is **identical before and after**. `dinofunk` renders its five voice rows (bass/bd/sd/hh/dino) with marks on both builds. No blank lane, nothing gone dark, zero page errors. The "0 net opaque nodes" number from the test pass now has a picture behind it.

Two oddities turned up and both are **pre-existing**, reproducing identically on the parent — flagging them here only so they aren't later misattributed to this PR:

- `arpoon` renders **0 lanes**
- `flatrave` has a voice row labelled with a NUL character

### The comma fix, visible on screen

The most interesting part. Lane count now tracks top-level comma arity:

| pattern | parent | this PR |
|---|---|---|
| `s("bd sd hh")` (no comma) | 1 | 1 |
| `s("[bd,cp]")` (bracketed) | **2** | **2** |
| `s("bd, cp")` | 1 | **2** |
| `s("bd, cp, hh")` | 1 | **3** |
| `s("bd, cp, hh, sd")` | 1 | **4** |

The bracketed row is what makes this readable: `[bd,cp]` has *always* produced one lane per stack arm. This PR brings top-level commas onto that same path — so the extra lanes are exactly the "a chord was being flattened into a sequence" fix showing up in the UI, landing consistently with behaviour that already existed.

### One correction to my own first read

I initially took those new `d2`/`d3` lanes for a regression, because they draw as short opaque blocks with the marks staying in `d1`. Screenshotting `[bd,cp]` on the **parent** showed the same short opaque second block — so that rendering is pre-existing stack-arm behaviour, not something this PR introduced. Without that control arm I'd have filed a bug against this PR that wasn't real. Filing it separately instead, since this PR does make it visible on a lot more tunes.

### Scope note

Worth recording for whoever picks this up next: the Pattern tab and the timeline *marks* are both insensitive to this change — the notation grid has been krill-backed since #906 — so lane structure was the only surface where any of this was visible at all.

**Verdict: clean under observation.** 🟢

PR #956, opened 2026-07-20
✅ test(ir): pin the control-classification behaviours no metric can see

MrityunjayBhardwaj:
Closes #955. Part of #929, guard rail for #944.

> **Scope, stated up front.** Everything pinned here is an **internal** surface — none of it is currently observable by a user. The suite is still worth having (`toStrudel` is the oracle for several test suites, so the collapse will move snapshots someone must triage), but it should not be priced as a user-facing fix. Evidence in "What this is not" below.

## Problem

The coming FX→Param collapse moves ~54 corpus nodes off the `FX` tag. Tag counts and parity snapshots will move loudly and review well — but several behaviours sit in blind spots *behind* those counts, where a change produces no count movement, no snapshot churn, no error and no failing gate.

Three independent reasons a blind spot exists:

1. **Zero corpus occurrences.** `reverb` and `hpf` appear in none of the 57 corpus files. `reverb` is not a core control (`isControlName('reverb') === false`), so `.reverb("0.3 0.5")` opaques while `.reverb(0.5)` classifies — the asymmetry a wholesale delete would erase. `scale` is the *other* non-registry control and behaves **differently**: it is a curated `Param` arm, so it classifies where `reverb` opaques. Reasoning about "the non-core controls" as one group gets one wrong, and neither is in the corpus to catch it.
2. **The metric is the wrong shape.** `extractTransform` (`toStrudel.ts:347`) matches `tag === 'FX'` to rebuild a transform arrow. A change there alters the emitted **string** while leaving every tag count identical.
3. **Semantics do not move either.** The two arrow forms are hap-equivalent, so even a behavioural test stays green.

## What this is not

Worth being precise, because an earlier draft of this PR overclaimed and the correction is instructive:

- **`toStrudel` has zero production callers.** The two modules that could call it document that they deliberately do not — `visualEdit/writeback.ts:11` and `visualEdit/arrange/serialize.ts:7` both explain that write-back does span surgery on **text** precisely to avoid this whole-statement regenerator, and `dist/index.d.ts:156` states it outright: *"NEVER calls `toStrudel` (no fidelity tax)"*. Neither arrow spelling reaches a user's document.
- **The split param keys are read by nothing but debug views.** `IRInspectorPanel.tsx:469` and `IRInspectorChrome.ts:29` call `Object.keys()`. Pitch resolution reads only `note`/`n`/`freq` (`musicalTimeline/pitch.ts`); audio gets `room`/`cutoff`/`delay` from Strudel's own eval, never from our IR.
- **The `() =>` form is the norm, not a fallback from a better path.** Of the corpus's 18 transform-arrow bodies, `add` (12) and `rev` (1) tag `Code`, `speed` (5) tags `Param`, `ply` (1) tags `Ply` — **all 18 already emit it**. The `x =>` FX form is the minority case.

So the collapse this guards is **internal coherence and drift removal**: it retires the last hand-maintained mirror of Strudel's control vocabulary. That matters on this repo's own record — chain roots (#953), mini grammar (#943), euclid helpers (#943) and this very control list (#935) each drifted and were silently wrong. It is a real reason to do the work, and not a reason to put it ahead of user-visible defects like #950.

## Fix

A guard suite pinning each behaviour **as it behaves today**, asserting on the **emitted string** and **collected event params** rather than tags. Every pin declares whether the collapse should *preserve* or *flip* it — an undeclared pin is a tripwire, and the next person deletes it.

| # | behaviour | collapse |
|---|---|---|
| 1 | `FX` body → `x => x.room(0.8)`; `Param`/`Code`/`Ply` → 0-arg arrow duplicating the receiver | **flips** |
| 2 | both arrow forms are hap-identical (eval oracle, 4 cycles) | preserves |
| 3 | `reverb` classifies numerically, opaques on a pattern arg; `scale` classifies; `hpf` classifies in both | preserves |
| 4 | `delay` splits the **tag** not the key; `lpf`/`hpf` split the **key** | **flips the numeric row** |
| 5 | `.room(0.3).room(0.5)` → `room: 0.3` (first-wins) | preserves |
| 6 | `jux` synthesises two `FX` pans (±1), no `userMethod`; a user `.pan()` is `Param` | preserves |

## Beliefs corrected by observation

Every assertion was probed against unmodified `f6c5049f` before being written. Three did not survive, and all three came from taking a prior framing on trust:

- **"Broken code" was wrong.** The hap oracle shows both arrow forms play identically — the 0-arg arrow ignores its argument and rebuilds the receiver, which is what the receiver was anyway.
- **"Degraded" was also wrong**, and backwards: the `() =>` form is what 4 of 5 tags and all 18 corpus arrows already emit. Retired the word.
- **A user-written `.pan(0.5)` is not `FX` at all** — it tags `Param`, so user and jux-synthetic pans are *already* on different tags, which is what makes the eventual tag deletion tractable. (My first explanation of *why* was also wrong — see the correction comment below.)

## Test plan

- `pnpm gate:editing:editor` — **3107 pass** (3080 + 27 added).
- `pnpm gate:editing:app` — **161 pass, unchanged**; no snapshot movement, confirming behaviour-neutrality.
- **Verified RED twice**, per the rule that a guard which cannot fail is worthless. First with the `Param` branch the collapse will add (6 pins red, in the intended places). Then with a sharper, more realistic fault — removing `delay` from the curated arms, a one-control mini-collapse — which turns the `delay` pin red **and only that pin** (26 unaffected). The pins are targeted, not blanket. Both faults reverted; **zero production changes** in this PR.
- `gate:editing:browser` not run: one unit test file, no production code, so the browser arm cannot be affected.

## Known gaps, on the record

- **App-side FX consumers have no pins.** `irProjection.ts` (5 sites) and `IRInspectorChrome.ts` (2) read the `FX` tag; the jux-pan strip predicate at `irProjection.ts:236` is pinned here only at IR level. That belongs with the tag deletion, since that is the change which moves those sites.
- **`room`'s key is unpinned** (only its first-wins merge is) — it keys to itself and is the largest predicted movement (+27), so the count prediction covers it. These pins are for what counts *cannot* see.

### DIFF
--- a/packages/editor/src/ir/__tests__/controlClassificationGuards.test.ts
+++ b/packages/editor/src/ir/__tests__/controlClassificationGuards.test.ts
@@ -16,14 +16,42 @@
  *      arrow; a change there alters the emitted STRING while leaving every tag
  *      count identical. Counting nodes cannot see a change in how a node is
  *      spelled.
- *   3. SEMANTICS DO NOT MOVE EITHER. The degraded arrow form is hap-equivalent
- *      to the idiomatic one (pinned below with the eval oracle), so even a
- *      behavioural test would stay green. Only the emitted string shows it.
+ *   3. SEMANTICS DO NOT MOVE EITHER. The two arrow forms are hap-equivalent
+ *      (pinned below with the eval oracle), so even a behavioural test would
+ *      stay green. Only the emitted string shows a change.
  *
  * So the assertions here are deliberately on the EMITTED STRING and the
  * COLLECTED EVENT PARAMS, not on tags. Each pin says whether the collapse is
  * expected to PRESERVE it or deliberately FLIP it — a pin whose intent is
  * unstated is just a tripwire, and the next person deletes it.
+ *
+ * ---------------------------------------------------------------------------
+ * SCOPE — what these pins are, and what they are NOT
+ * ---------------------------------------------------------------------------
+ * Read before treating a failure here as a user-facing bug. It is not one.
+ * Both surfaces this file pins are INTERNAL:
+ *
+ *   - `toStrudel` has ZERO production callers. The two modules that could call
+ *     it document that they deliberately do not (`visualEdit/writeback.ts:11`,
+ *     `visualEdit/arrange/serialize.ts:7`): write-back does span surgery on
+ *     TEXT precisely to avoid this whole-statement regenerator, and
+ *     `dist/index.d.ts:156` states the reason outright — "NEVER calls
+ *     `toStrudel` (no fidelity tax)". So neither arrow spelling ever reaches a
+ *     user's document. `toStrudel` IS the oracle for several test suites, and
+ *     that is the reason to pin it.
+ *   - The split param keys (`params.lpf` vs `params.cutoff`) are read in
+ *     production by NOTHING except the IR Inspector's `Object.keys()` debug
+ *     views (`IRInspectorPanel.tsx:469`, `IRInspectorChrome.ts:29`). Pitch
+ *     resolution reads only `note`/`n`/`freq` (`musicalTimeline/pitch.ts`).
+ *     Audio gets its `room`/`cutoff`/`delay` from Strudel's own eval, never
+ *     from our IR.
+ *
+ * The collapse these pins guard is therefore INTERNAL COHERENCE and drift
+ * removal — retiring the last hand-maintained mirror of Strudel's control
+ * vocabulary — not a fix for anything a user can currently observe. That is a
+ * good reason to do it, and a bad reason to rush it ahead of defects that ARE
+ * user-visible. Stating the scope here so the next reader prices it correctly
+ * rather than inferring urgency from the density of assertions.
  *
  * Everything asserted here was verified against unmodified `f6c5049f` before
  * being written down. Nothing here is transcribed from a design document.
@@ -92,21 +120,23 @@ async function haps(code: string, cycles = 4): Promise<string[]> {
 // ---------------------------------------------------------------------------
 // 1. extractTransform — the FX-only privileged path (toStrudel.ts:347)
 // ---------------------------------------------------------------------------
-describe('extractTransform: FX is the exception, not the rule', () => {
+describe('extractTransform: the FX arrow form is the minority path', () => {
   // COLLAPSE WILL FLIP THIS. `x => x.room(0.8)` is reachable ONLY because a
   // numeric-arg control still tags FX. Once room/delay/lpf become Param, this
   // branch has no real users left and the emission becomes the `() =>` form
   // below. That flip is EXPECTED and correct-to-make; it must be DECLARED.
-  it('an FX body emits the idiomatic single-argument arrow', () => {
+  it('an FX body emits the single-argument arrow form', () => {
     expect(emit('s("bd hh").every(4, x => x.room(0.8))'))
       .toBe('s("bd hh").every(4, x => x.room(0.8))')
   })
 
   // COLLAPSE PRESERVES THESE. They already take the generic fallback today —
-  // which is the point: the degraded form is the MAJORITY behaviour, covering
+  // which is the point: the `() =>` form is the MAJORITY behaviour, covering
   // all 18 transform-arrow bodies in the corpus (add 12, speed 5, rev 1, ply 1).
   // The fallback emits a ZERO-ARGUMENT arrow that inlines a duplicate copy of
-  // the receiver. It discards its input entirely.
+  // the receiver, discarding its input entirely — which is verbose but, per the
+  // hap pins below, plays identically. Calling it "the norm" is not a value
+  // judgement either way; it is simply which path most bodies take.
   it.each([
     // Every corpus transform-arrow shape, with the tag it carries. The tag is
     // ASSERTED, not just documented — otherwise this table would keep passing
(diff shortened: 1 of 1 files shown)


MrityunjayBhardwaj:
**Self-review pass — three gaps closed** (`b8fba79d`). Each was in exactly the blind-spot class this suite exists to cover, which is a fair sign the first pass was not paranoid enough.

- **`delay` was unpinned** — the flagship case (16 FX + 3 Param in the same corpus). It also turns out to have a *different shape* from `lpf`/`hpf`: it splits the **tag** without splitting the **key**, because `getControlName('delay')` is already `delay`. So "did the key move?" is not a sufficient check — for `delay` only the tag shows the defect, for `lpf` only the key does. Both shapes are now pinned.
- **`scale` was unpinned**, though #944 names it beside `reverb` as the other non-registry control. They do **not** behave alike: `reverb` is a curated **FX** arm so its pattern form opaques; `scale` is a curated **Param** arm so it classifies. A collapse that reasons about "the non-core controls" as one group gets one of them wrong — and neither appears in the corpus to catch it.
- **The transform-arrow table carried a tag column it never asserted**, so it would have kept passing if a body changed tag while its emission happened to still match. Now asserted.

Also added the remaining curated arms (`crush`/`begin`/`resonance`/`cut`) as a set, so the collapse cannot quietly drop one.

**Verified RED again**, with a sharper fault than the first round: removing `delay` from the curated arms — a one-control mini-collapse, the most realistic possible defect — turns the `delay` pin red **and only that pin** (26 others unaffected). The pins are targeted, not blanket.

Still zero production changes. Editor **3107** pass, parity **161** unchanged.

---

**Known gaps I am deliberately leaving out of this PR**, so they are on the record rather than implied-covered:

- **The app-side FX consumers have no pins.** `irProjection.ts` (5 sites) and `IRInspectorChrome.ts` (2 sites) read the `FX` tag, and the jux-pan strip predicate at `irProjection.ts:236` is pinned here only at the IR level, not at the projection level where it is actually consumed. That belongs with the tag deletion, since that is the change which will move those sites.
- **`room`'s key is unpinned** (only its first-wins merge is). `room` keys to itself and is the largest predicted movement (+27), so it is well covered by the count prediction — the pins here are for what counts *cannot* see.

MrityunjayBhardwaj:
**Reframing pass** (`f68add2b`) — the prose in this PR, the issue, and the suite's own header were overclaiming. Correcting on the record, since the correction is more useful than the original claim.

I had described one emission form as "degraded" and the split param keys as "the defect at the consumer-facing level." I then went looking for the consumers, and neither claim survived:

- **`toStrudel` has zero production callers.** The two modules that could call it carry comments explaining that they deliberately don't (`writeback.ts:11`, `arrange/serialize.ts:7`), and `dist/index.d.ts:156` says it outright: *"NEVER calls `toStrudel` (no fidelity tax)."* Write-back does span surgery on text specifically to avoid this regenerator. So the "source rewriting" I used to justify the word never reaches a user's file.
- **The split keys are read only by debug views** — `Object.keys()` in the IR Inspector. Pitch reads `note`/`n`/`freq`; audio gets its values from Strudel's eval.
- **"Degraded" had the polarity backwards.** The `() =>` form is what 4 of 5 tags and all 18 corpus transform-arrows already emit. It is the norm; the `x =>` FX form is the minority path.

Provenance of the error, since it is the reusable part: the handoff into this session said "emits BROKEN CODE." I disproved that with the hap oracle — and then **softened the word to "degraded" instead of retiring it**, carrying the original framing's urgency into a claim I had just falsified. The check I skipped was the cheap one: *who reads this output?* Two greps.

**What changed:** a SCOPE section in the suite header with the file:line evidence, "degraded" retired throughout for neutral description, and the key-split re-framed as internal incoherence — one control answering to two keys misleads the next person reasoning about the IR, which is a real cost and a different one from anything a user can observe.

**What did not change:** the tests. All 27 still pass, still verified RED. They pin `toStrudel`, which *is* the oracle for several suites, so the collapse will move snapshots and someone has to triage intent. The suite was right; its billing was wrong.

**Consequence for sequencing.** The collapse is architectural hygiene — retiring the last hand-maintained mirror of Strudel's vocabulary, which this repo's record justifies (#953, #943, #935 all drifted silently). It is *not* a user-facing fix, and on that basis it probably ranks below **#950** (a stack arm's lane renders empty, with marks folded into the first lane) — a defect users can actually see. Flagging for your call; I have not reordered anything.

MrityunjayBhardwaj:
**Correction** (`35671d76`) — and it came out of rewriting #944, which is worth recording because the two are connected.

#944 cites six line numbers. Per the rule that a cited line must be re-grepped rather than trusted, I checked all six before reusing them — **three had drifted**: the FX arms `2340`→**2376**, `CHAIN_ROOT_RECOGNISER` `995`→**998**, the registry `default` arm `2584`→**2611**. Reading the real lines is what exposed my own error.

**I claimed `pan` is not a curated arm.** It is. The curated method switch has **two** groups, and `pan` sits in the **Param** one (`parseStrudel.ts:2575`, beside `s`/`n`/`note`/`scale`/`speed`) — not the FX one (`:2376`, `room`/`delay`/`lpf`…). 13 FX arms + 11 Param arms = the 24 curated arms the north star names. I had inferred "not in the FX list" ⇒ "not curated," and never looked at the second list.

The assertion was right and stayed green; only the explanation was wrong — which is the more dangerous half, because a reader trusts the comment to know *which mechanism* the collapse will move, and mine pointed at the registry when the answer is the curated switch. Fixed in the test comment, in #955, and in the PR body above.

Third time this session the same shape has bitten: **the claim was checkable in two greps, and I reasoned instead.** ("broken code" → hap oracle disproved it · "degraded" → zero production callers disproved it · "pan isn't curated" → line 2575 disproves it.) Each was caught only because something forced a second look. Worth a catalogue entry at wrap: *a cited line number and a stated mechanism are both claims, and neither inherits the credibility of the assertion it accompanies.*

---

**#944 is rewritten.** The old acceptance — "Code/Param/**FX counts unchanged**" — would have failed a correct implementation and passed a no-op, since the collapse must deliberately move ~54 nodes off `FX`. It now carries the falsifiable prediction captured before any change (`FX 60→6 · Param 239→293 · Code 181→181`, with per-key movement summing to exactly +54), explicit abort conditions, verified line citations for all five curated sets, and the `reverb`/`scale` asymmetry that makes a wholesale delete unsafe. The `FX` tag deletion is split out as a separate follow-on so this prediction stays falsifiable.

PR #962, opened 2026-07-20
🔬 chore(ir): publish the predicate audit for parseStrudel.ts

MrityunjayBhardwaj:
Part of #929 / #925. Implements the deliverable in #959 — a table, not a refactor.

## The problem

`parseStrudel.ts` decides questions about JavaScript syntax and about Strudel's vocabulary in **42 hand-written anchored regular expressions**. Every other module in the parse path asks somebody who already knows, and has none:

| module | asks | anchored regexes | lines |
|---|---|---|---|
| `ir/parseMini.ts` | krill | **0** | 397 |
| `ir/collect.ts` | — | **0** | 1419 |
| `ir/parseStrudelStages.ts` | — | **0** | 377 |
| `visualEdit/chunkDetect.ts` | acorn | **0** | 496 |
| `visualEdit/arrange/parse.ts` | acorn | **0** | 223 |
| **`ir/parseStrudel.ts`** | **nobody** | **42** | **3335** |

A regular expression cannot express an infinite grammar, so each of the 42 is incomplete by construction — and none of them throw. That is why these are only ever found by being hit, one incident at a time.

## What the audit found

The 42 answer only ten distinct questions. Six of them are JavaScript's, and one — "is this a string literal, and what is inside it?" — is transcribed **fifteen times in six different spellings**. The `note`/`s`/`mini` cluster is ten regexes because three name-groups were each spelled once per quote style: a product of two independent transcriptions.

The failure modes are graded, and the worst one is invisible:

1. **The node opaques** — keeps its bytes, loses its structure. Honest, and visible in the parity counts.
2. **The node changes meaning** — `.s("drum/kit")` stops being a sample name and is re-read as mini-notation.
3. **The node silently disappears** — `.every(2, fast(2*2))` yields an `Every` wrapping the *untransformed* pattern. No `Code` marker, no count movement, no error. The IR asserts a transform the source does not contain, and the view draws it.

Class 3 is the argument for counting these rather than fixing them one at a time.

## Observed, not argued

Eleven incompleteness claims were run through `parseStrudel` against a control arm differing only in the property under test, and the IR tag-paths compared. A sample:

| input | control | result |
|---|---|---|
| `const café = …` | `cafe` | whole program opaques |
| `pp => pp.fast(2)` | `x => x.fast(2)` | **`Fast` node vanishes** |
| `fast(2*2)`, `slow(-2)` | `fast(2)`, `slow(2)` | **transform vanishes** |
| `.mask('1 0')` | `.mask("1 0")` | opaques |
| `.s("b\"d")` | `.s("bd")` | opaques |
| `setGain(0.5)` | `setcps(0.5)` | whole program opaques |
| `/*x*/stack(…)` | `stack(…)` | whole program opaques |
| `.s("drum/kit")` | `.s("bd")` | re-read as mini-notation |

Claims that were only reasoned from the expression are marked as such, and not mixed in with these.

Two probes came back SAME and are reported that way: the numeric arm of `classifyLiteralRhs` is confirmed still closed after the leading-decimal fix, and one early probe never reached `parseTransform` at all — it was the probe that was wrong, not the predicate.

## The gate

`predicateAudit.test.ts` re-derives the census from the source on every run and fails in **both** directions: a new predicate with no entry, and an entry the source no longer has. A census that overstates reads exactly like one that is current.

Two details worth the review:

- It asks **TypeScript's own tokenizer** which slashes are regexes rather than scanning for them. Deciding that means knowing which slashes are division, which are in strings and which are in comments — the same class of question the audit is about, so hand-rolling it here would reproduce the defect inside the instrument that measures it. It also makes the exclusion of the four regexes quoted in prose structural rather than heuristic.
- The five-module zero is a negative result, so both the document and the gate carry a **control arm** — the same scan must find the known 42 in `parseStrudel.ts`. A zero from a broken query would otherwise read as a delegating module.

Verified red by injecting a new anchored regex (fails), and by deleting an entry that is still in the source (fails). Green after restore.

## Scope

No behaviour change: one document, one test. `parseStrudel.ts` is untouched, so no `dist` rebuild.

The sequencing stays in #959 — category B is the same shape as the chain-root work and needs no new dependency; category A goes behind a spike, because the honest opening question is whether the IR needs its own JavaScript parse at all when `visualEdit/` already has one.

## Test plan

- `pnpm gate:editing:editor` — 3183/3183 (203 files)
- `pnpm gate:editing:app` — 161/161 (11 files)
- `tsc --noEmit` — 73 pre-existing, **0** in the new file
- fault injection — gate red on a new regex, red on a deleted entry, green on restore

Browser arm not run: this change touches no runtime code, no `dist`, and no rendered surface.

### DIFF
--- a/packages/editor/src/ir/PREDICATE-AUDIT.md
+++ b/packages/editor/src/ir/PREDICATE-AUDIT.md
@@ -1,0 +1,336 @@
+# Predicate audit — `ir/parseStrudel.ts`
+
+Every anchored regular expression in `parseStrudel.ts`, the question it answers, and the
+system that owns the right answer.
+
+This file is a **census, not a plan**. It exists so that "find the next parser bug" becomes
+"close a finite list". Nothing here changes behaviour, and no entry is an instruction to
+refactor — the sequencing lives in the issue.
+
+`predicateAudit.test.ts` re-derives the census from the source on every run and fails if the
+two disagree, so a new predicate cannot be added without appearing here first.
+
+---
+
+## Why this file exists
+
+`parseStrudel.ts` decides things about JavaScript source text and about Strudel's vocabulary.
+It answers those questions itself, in hand-written regular expressions. Every other module in
+the parse path asks somebody who already knows:
+
+| module | asks | anchored regexes | lines |
+|---|---|---|---|
+| `ir/parseMini.ts` | krill | **0** | 397 |
+| `ir/collect.ts` | — | **0** | 1419 |
+| `ir/parseStrudelStages.ts` | — | **0** | 377 |
+| `visualEdit/chunkDetect.ts` | acorn | **0** | 496 |
+| `visualEdit/arrange/parse.ts` | acorn | **0** | 223 |
+| **`ir/parseStrudel.ts`** | **nobody** | **42** | **3335** |
+
+Every module that delegates has zero. The one that does not has forty-two. `parseMini.ts` is
+the controlled before/after: 512 lines with a hand-rolled tokenizer, 397 lines and no anchored
+regexes after it was rebuilt on krill.
+
+(The five zeroes are a negative result, so they carry a control arm: the same query finds
+anchored regexes in fifteen other `.ts` files in this package, and `parseMini.ts` retains two
+*unanchored* single-character scans — `/[0-9.]/` and `/\s/` at `parseMini.ts:302,377` — which
+locate a boundary rather than decide what a token means.)
+
+### The failure mode is silence
+
+A regular expression cannot express an infinite grammar, so every transcription here is
+incomplete by construction. None of them throw. Three things happen instead, in increasing
+order of harm:
+
+1. **The node opaques.** It becomes `Code`, keeps its bytes, and loses its structure. Honest,
+   and visible in the parity counts.
+2. **The node changes meaning.** `.s("drum/kit")` stops being a literal sample name and is
+   re-read as mini-notation.
+3. **The node silently disappears.** `.every(2, fast(2*2))` produces an `Every` wrapping the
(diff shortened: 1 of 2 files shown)


MrityunjayBhardwaj:
## Self-review — the gate had the blind spot the audit is about

Second commit (`3c02d0e9`). Three defects in my first pass, all the same shape: a claim recorded without being measured.

**1. The gate could not see two of the predicates it was auditing.** It filtered to *anchored* regexes, which excluded `ARITH_SPLIT` — a transcription of JavaScript's arithmetic operator set that happens to carry no `^` or `$`. A census that decides for itself what counts as a predicate reproduces the exact failure mode it exists to catch.

Fixed by removing the judgement from the gate: it now counts **every** regex literal (49), and the anchored/unanchored classification lives in the document where it can be disagreed with. New **category D** holds the seven unanchored — `ARITH_SPLIT` and the block-comment matcher as genuine predicates with owners, five character scans as boundary-locators with none. Red-tested by injecting an unanchored regex, the case the first version missed.

**2. "Fifteen other `.ts` files" was read off a truncated listing.** The real count is 42. The control arm intended to make the five-module zero trustworthy was itself an unverified number.

**3. The Category C sites were quoted from the issue, not looked up.** Two of three line numbers pointed at unrelated code; `CHAIN_ROOT_RECOGNISER` is at `:1039`. Now verified against the source, and all line numbers are marked indicative — the gate keys on regex source precisely so they can drift without reading as new predicates.

Worth recording how #3 surfaced: a grep for `CHAIN_ROOT_RECOGNISER\s*=` returned zero while line 1959 *uses* the identifier. The query was wrong, not the codebase — the same failure that produced the truncated count, one command earlier.

Gates re-run after the fixes: editor 3183/3183, tsc 73 pre-existing / 0 new.

PR #964, opened 2026-07-20
🐛 fix(timeline): a stack arm's lane shows its own notes

MrityunjayBhardwaj:
Closes #950 (close by hand — `Closes` does not fire from a studio base).

## Problem

`$: s("bd, cp")` draws two lanes and fills only the first. Every mark piles onto `d1`; `d2` renders as an empty band. It **plays correctly** — the pattern is fine, the display is not.

Pre-existing: bracketed `[bd,cp]` has always rendered this way. The krill adapter made it *common* rather than caused it — top-level commas are no longer flattened, so every `$:` whose mini contains a comma now grows one lane per arm, and each new arm inherited the empty rendering. `s("bd*2, ~ sd, hh*4")` showed one populated lane and two blanks.

## Diagnosis

The issue guessed this was the eval-backed-lanes seam and might already be closed by the declared-track anchor fix. **It is not**, and seeding more anchors cannot fix it.

Measured in the running app, broken case against the known-good two-track control:

| | `$: s("bd, cp")` | two real `$:` tracks |
|---|---|---|
| `analysis.lanes` | `["d1","d2"]` | `["d1","d2"]` |
| `marksByLane` | **`[["d1", 8]]`** — d2 empty | `[["d1",4],["d2",4]]` |
| `labelOffsetByLane` | **`[]`** | `[["d1",0],["d2",11]]` |
| clips | both `0→4` | both `0→4` |

**Lanes and marks are partitioned by different keys.** Lanes come from the IR structure — one per stack arm. Marks are attributed by source containment: each hap goes to the lane whose `Track` offset is the largest at or before the hap's own offset. Arms from a real `$:` statement carry that offset; arms from expanding a comma inside ONE statement had none. The anchor map came out empty, every hap fell back to the engine trackId — identical for all arms of one statement — and the two partitions disagreed.

Two corrections to the issue text, both from the dump: the clip is **not** short (d2 spans the full `0→4` like every lane — the "block spanning cycles 0–1" was my misread of a screenshot; an empty lane just draws as a flat band), and the anchor map is **empty**, not merely incomplete.

## Fix

Anchor a comma-expanded arm on **its own source span**. The information was already present on both sides — `bd` at 6, `cp` at 10 — only the key was too coarse. Containment then partitions the mini string *by arm*, which is the same rule the statement case already uses, one granularity finer. Statement offsets still win when present, so the real-`$:` path is byte-identical.

**The span is the minimum start / maximum end over the arm's subtree, not the top node's own `loc`.** Two shapes make the top node wrong, and I got this wrong on the first pass:

- a combinator's `loc` covers its **operator**: `bd*2` reports 8 while the `bd` it plays is at 6. Anchoring there lands *after* the haps it must catch and reproduces the very fold being fixed.
- a multi-element arm has no `loc` at all: `~ sd` is a `Seq` with `loc: undefined` over located children.

Both were caught by the gate going red, not by reading the code.

## Observed

Real app, fresh build, after rebuilding the editor `dist` the app actually serves:

`$: s("bd*2, ~ sd, hh*4")` now draws three lanes each with its own rhythm — the middle one visibly offset, a rest then a snare — where it drew one populated lane and two empty bands. `s("[bd,cp]")` and `s("bd, cp")` render equivalently, as the issue requires.

## On the gates

**Deliberately no browser spec.** Lane *count* never discriminated here: both lanes appeared before the fix, one of them blank, so a count assertion passes either way — it would have been a gate that cannot fail. Marks are canvas, not DOM, so there is nothing assertable at that level.

Instead two unit gates, one on the cause and one on the effect:

- `parseStrudelStages.armLoc.test.ts` — each arm gets a distinct, ascending anchor, through the **real** staged pipeline. Includes the statement-anchored cases as controls.
- `collectNoteMarks.stackArms.test.ts` — marks actually split across arms. Mocks `collectCycles` to return *nothing* (which is what forces the wrapper path this bug lives on) and builds the IR by running the real pipeline over real source, so the anchors under test are the ones the app gets.

Verified red without the fix: editor 3 of 5 fail — and the 2 that stay green are the statement-anchored controls, which is what makes them controls — app 4 of 4 fail. Green with it.

## Test plan

- editor **3184/3184** (203 files)
- app **767/767** (56 files)
- `gate:editing:app` **161/161** — no downstream snapshot moved
- `tsc --noEmit` — 73 pre-existing, 0 new
- editor `dist` rebuilt and committed whole (the app serves dist)


MrityunjayBhardwaj:
## Self-review

**The gate caught a real defect in my own first attempt.** I initially anchored on the arm's top node `loc`. That works only for a bare single-atom arm, and the three-arm case went red immediately: `bd*2` reported the `Fast` operator at 8 while the `bd` it plays is at 6 — anchoring *after* the haps it must catch, which reproduces the exact fold being fixed — and `~ sd` is a `Seq` with no `loc` at all. The subtree min/max is the second attempt. Worth stating plainly because the first version would have looked correct on the issue's own example.

**The overlap risk, checked rather than assumed.** The anchor is the extent of the arm's subtree, so anything leaking into that subtree widens the arm — and a chain applied to the whole stack (`s("bd, cp").gain(0.5)`) is both the commonest real-world shape and the one that would do it. If `.gain`'s loc reached the arms, both would extend to the chain call, their spans would overlap, and the ordering containment depends on would break. Measured across `.gain`, `.slow().room()`, a nested `[cp, hh]` arm and a `.scale().s()` chain: **no overlap** — chain methods apply to the stack, not per-arm. Pinned in `7919679e`, since nothing else would notice it changing.

**Known residual, not fixed here.** A hap whose `loc[0]` pointed *past* every arm (at a chain call rather than a mini atom) would now attribute to the last arm instead of falling back to trackId. Observed `loc[0]` is the innermost mini atom in every case measured — 6 and 10 for `s("bd, cp")` — so this is latent, not live. Recording it rather than guarding speculatively.

**What I'd still call thin.** The haps in the app-side gate are synthetic; only their offsets are real. The end-to-end path is covered by observation (screenshots on a fresh build) rather than by an assertion, because marks are canvas and lane count does not discriminate. That is a genuine gap in the automated coverage, and the reason the observation was done on the real app rather than inferred.

Gates after the follow-up commit: editor **3185/3185**, app **767/767**, parity **161/161**, tsc 73 pre-existing / 0 new.

Issue #132, opened 2026-05-15
parseStrudel: recursive parsing inside note/n/s args (nested mini + chain)


MrityunjayBhardwaj:
## Problem

`parseStrudel` falls back to opaque `Code(<verbatim>)` for expressions like:

\`\`\`
n(\"[0,3] 2 [1,3] 2\".fast(3).lastOf(4, fast(2))).clip(2).offset(...)
\`\`\`

The OUTER chain (\`.clip\`, \`.offset\`) now walks correctly (Phase 20-14 γ
cluster-B multi-line fix). But the ROOT \`n(...)\` arm requires its
argument to be a bare \`\"...\"\` quoted string per regex
\`/^(?:note|n)\\s*\\(\\s*\"([^\"]*)\"\\s*\\)/\`. When the inner argument is itself
a mini-string with its own method chain (\`\"...\".fast(3).lastOf(...)\`), the
regex fails and the whole expression becomes opaque Code.

## Why it matters

Real strudel.cc programs commonly nest expressions inside \`n()\`/\`note()\`.
The \`arpoon\` tune in our parity corpus is the surfaced case (Phase 20-14
γ-wave). \`barryHarris\` partially affected; others may surface as the corpus
grows.

In Stave today this means: any pattern that uses \`n(MINI.method().method())\`
loses IR structure → loses click-to-source, loses Inspector projection,
loses per-event chrome on the timeline. The audible result is correct (the
transpiler reifies the inner string and the outer methods chain through
the runtime), but the IR is opaque.

## Proposed approach

Replace the regex-string arg with recursive expression parsing. In \`parseRoot\`
(or whatever dispatches \`note\`/\`n\`/\`s\`/etc.):

1. Match the OUTER call shape \`^(?:note|n)\\s*\\(\` (open paren only).
2. Find the matching close paren via brace-depth tracking (same algorithm
   stripParserPrelude already uses for skipping multi-line \`samples({...})\`).
3. Pass the inner-arg substring to \`parseExpression\` recursively.
4. If the recursive parse returns a typed IR, wrap it in the appropriate
   note/n constructor (likely \`IR.note(parsed)\` or similar). If it returns
   Code-fallback, fall back to the existing opaque wrap (current behavior).

The recursive call MUST pass through the absolute offset so loc stays
correct (additive offset arithmetic — same pattern as cluster-B's chain walker).

## Acceptance

[...]

PR #965, opened 2026-07-21
♻️ refactor(ir): delegate the pattern-source root grid to acorn

MrityunjayBhardwaj:
Part of #959 / #929 (north-star invariant #2 — classification delegates, it does not hand-roll). The first cut into `parseStrudel.ts`'s 42 regexes, taken on the biggest cluster.

## What collapsed

`note`/`n`/`s`/`sound`/`mini` on a string argument was **nine regexes** — three name-groups × three quote styles:

```
note/n   × " ' `   ┐
s/sound  × " ' `   ├ 9 regexes, two orthogonal questions asked as a product
mini     × " ' `   ┘
```

The audit (#962) measured this as the largest cluster in the file. Each regex fused two independent decisions — the **name** (sample-vs-note semantics) and the **delimiter** (`backtickInnerToIR` vs `parseMini`; only a backtick spans newlines) — and each recomputed the inner mini's source offset by hand.

`extractPatternSourceCall` replaces all nine with one delegation to **acorn**, which owns "is this a string literal, and where does it start". The two real decisions (name → sample flag, delimiter → backtick flag) are all that remain, over one shared read. A cheap five-name gate stays in front so a non-source root (`stack(...)`) is not parsed just to be declined, and the loose recursive arm for a chained inner (#132) is untouched.

## Why it is safe — spiked before written

This is the first cut into a 3335-line file with hard loc-fidelity constraints, so I spiked it before touching anything:

- **Offset-safe.** Over 58 inputs (hand cases + real corpus root calls), acorn reproduced the exact `(isSample, isBacktick, inner, quoteOffset)` the regexes produce — **zero mismatches**. `arg.start` in trimmed-coordinates plugs into the identical `baseOffset + leadingWs + … + 1` math.
- **The one real risk, checked not assumed.** The old regexes are not `$`-anchored, so they match a *prefix* — a chained root `note("c3").fast(2)` would have them silently match `note("c3")` and drop the chain, where acorn (a MemberExpression callee) declines. I confirmed against the real parser that **the chain survives** (`note("c3").fast(2)` → a `Fast` node), which proves `parseExpression` splits the chain before `parseRoot`. So a chained root never reaches these arms; acorn and the regexes agree on every input they actually receive.
- **Behaviour-preserving, not a reach change.** The spike's "acorn accepts more" list was empty — the regexes' `\s*` already handle whitespace, and the theoretical incompleteness (escaped quotes) doesn't occur in mini-notation. Six edge cases captured before the change (including `` note(`${x}`) `` and `s("b\"d")`) round-trip **byte-identical** after it — the escaped-quote case reaches `parseMini`'s own fallback and still lands on Code.

## The proof

- **Parity corpus moves no snapshot** in either package — the real evidence over the whole corpus (editor mini-corpus + round-trip; app `parity-corpus` 161/161).
- **Observed in the real app** on a fresh build: `note`/`s`/`mini` roots render three lanes with a pitched melody and no errors, confirming acorn runs cleanly in the **browser bundle** (it already ran in three `visualEdit` modules).

## Gates

- editor **3197/3197** (205 files) — includes a new `parseStrudel.patternSource.test.ts` pinning all nine old shapes, the sample/note/mini semantics, offset fidelity (red-tested with an injected off-by-one), and the declines (chained inner, `${}` template, non-source callee)
- app `gate:editing:app` **161/161** — no downstream snapshot moved
- `tsc --noEmit` — 73 pre-existing, **0** new
- editor `dist` rebuilt and committed whole (the app serves dist; no `.d.ts` change — the helper is internal)

## Audit

`PREDICATE-AUDIT.md`: **42 → 34** anchored regexes; B1 goes from ten sites to two (the name gate + the loose arm), the extraction recorded as delegated. The `predicateAudit.test.ts` gate enforced the update — it went red on the count until the doc matched.

The name vocabulary is still a hand-list of five; delegating *that* to `controls.mjs` is #944's job, not this collapse.


MrityunjayBhardwaj:
## Self-review

**Old vs new diffed on 13 non-corpus edge cases — byte-identical.** The parity corpus proves no snapshot moved, but the corpus doesn't exercise the awkward declines, so I checked out the parent's `parseStrudel.ts` and ran both against the same inputs. Identical IR tags AND round-trip for every one:

| input | both produce | the risk it covers |
|---|---|---|
| `notes("c3")` | Code | superset name must NOT match `note` (gate anchors at `^`, acorn checks the exact name) |
| `n(42)` | Code | numeric arg declines (acorn: `Literal` but not a string) |
| `note("a\nb")` | Code | escaped newline — raw source slice, not the JS-unescaped value |
| `sound("bd")` | `s("bd")` | alias normalisation preserved |
| `s('x')` | `s("x")` | single quote → sample semantics |
| `note("c3").s("piano")` | Param+Play | chained root — the chain survives (split before parseRoot) |
| `note("")` | Pure | empty inner |
| `s()` | Code | zero args |

**The one behaviour I deliberately allowed to differ turned out not to.** `s("b\"d")` (escaped quote) — the old regex declined it to Code, acorn accepts it. But the inner `b\"d` isn't valid mini-notation, so `parseMini`'s own fallback returns Code anyway. Same outcome, reached one layer down. So even the "acorn is stricter/looser" case is invisible.

**Residual, recorded not guarded:** the five-name vocabulary is still a hand-list (`PATTERN_SOURCE_NAMES` + the gate regex). Delegating *that* to `controls.mjs` is #944's scope — this PR delegated the string/offset half, which is acorn's, and left the vocabulary half where the audit's category B already tracks it.

**What I'd flag as the honest cost:** acorn now parses every plausible source root (one small `.parse` per `note`/`s`/`mini` call), where before it was a regex match. For a song of tens of tracks this is negligible, and the five-name gate keeps `stack(...)`/bare roots off the acorn path entirely — but it is a real change to the parse hot path, not zero-cost.

Gates after review: editor 3197/3197, app parity 161/161, tsc 73 pre-existing / 0 new.

Issue #990, opened 2026-07-24
🧭 Phase 3f: refusals should name the gate that actually stopped them, not the first writer to decline


MrityunjayBhardwaj:
Follow-on to #986 (P2 of the leaf-loc write-back plan), and the first half of what a fresh measurement says we should do next.

## The problem

With P1a (grid) and P1b (roll) merged, three writers now stack behind one parse call:

```
parse{StepGrid,PianoRoll}Core   →   projection (element re-emit)   →   leaf projection
```

When a unit is finally refused, the `reason` string it reports is **whichever writer declined first** — almost always the syntactic core. Nothing downstream updates it. So the reason describes a subsystem that may have nothing to do with why the unit is actually unavailable.

This is no longer a tidiness problem. It is actively misdirecting our roadmap.

## The evidence

I instrumented the real projection functions to record which gate actually refuses each still-broken unit, and swept the full 1500-unit corpus. First failure wins, so the buckets are exclusive:

| gate | step | roll | total |
|---|---:|---:|---:|
| onset-null (values aren't this surface's) | 418 | 833 | **1251 (80%)** |
| unstable-period | 98 | 100 | **198 (13%)** |
| edit-safe probe | 40 | 1 | 41 |
| irrational onset | 23 | 2 | 25 |
| leaf anchors (the write-back guard) | 2 | 14 | 16 (1%) |
| view-unusable / silent / MAX_STEPS / mixed | 8 | 18 | 26 |

Now take the class our blocker backlog has ranked near the top for months — **nested groups**. Where those units actually die:

```
onset-null 31 · unstable-period 11 · irrational 5 · edit-safe 3 · leaf anchors 0
```

**Zero** of them are refused for anything to do with nesting. They report "nested groups are beyond the editable subset" because that is what the *core* said, while the real cause is that they are the wrong surface (a drum pattern asked of the piano roll) or that their period exceeds the projection cap.

We have been steering by a signal that points at the wrong subsystem.

## What to do

Replace the cluster of syntactic reason strings with the single rule all three writers already implement:

> A cell/note is editable **iff** it maps to exactly one source leaf, and an edit is a byte replacement at that leaf. Everything else is copied verbatim.

Concretely:

- **Report the final gate, not the first decline.** A refusal should name the gate that actually stopped it.
[...]

PR #992, opened 2026-07-24
🧭 feat(notation): refusals name the gate that actually stopped them

MrityunjayBhardwaj:
Phase 3f of the code↔view epic (#925), and the first of the sequence #990 → #991 → the bijection statement. Refs #990.

## The problem

Three writers stack behind one parse call:

```
parse{StepGrid,PianoRoll}Core   →   element projection   →   leaf projection
```

The `reason` a refused unit reported was whichever declined **first** — almost always the syntactic core. Nothing downstream updated it, so the reason named a subsystem that often had nothing to do with why the unit was unavailable.

The clearest case: every corpus unit whose refusal read *"nested groups are beyond the editable subset"*. Re-measured against the shipped parsers, all 51 of them die somewhere else:

```
wrong-surface 19 · no-note-content 12 · unstable-period 11 · irrational-onset 5 · edit-unsafe 3 · resolution 1
leaf write-back guard: 0
```

Not one is stopped by nesting. That class sat near the top of our blocker backlog for months.

## The change

Each projection now declines at a named gate instead of a bare `null`, and a refusal carries that gate plus a sentence derived from it. The gates are not a new feature list — each is one face of the editability invariant (a played onset is editable iff it maps to a unique disjoint source span, in a view that stays true): `wrong-surface`, `no-note-content`, `unstable-period`, `mixed-pitch-domain`, `irrational-onset`, `resolution`, `element-tiling`, `no-leaf-anchor`, `edit-unsafe`, `view-unusable`.

**Which writer's gate gets reported** turned out to matter, so I measured both rules rather than picking one:

| reported gate | result |
|---|---|
| deepest across both writers | 144 units blame `element-tiling` — but the element writer checks for a whole-cycle `<…>` *before it reads a single hap*, so 65 of those are really the wrong surface and 58 never repeat |
| the leaf projection's | reproduces the gate-attribution sweep this phase was planned from, exactly |

The leaf writer is the general write-back — it needs only that each played note own a disjoint source token — so its obstruction is the answer to "what would have to change". The element re-emit's difficulties are facts about a specialization. A gate that fires before the values are read is not progress.

## The anatomy, from the shipped parsers

```
===== REFUSAL GATES (1500 corpus units, both surfaces) =====
  refusals: 1557
  gate                  step  roll  total
  wrong-surface          344   692   1036  (67%)
  no-note-content         78   145    223  (14%)
  unstable-period         98   100    198  (13%)
  edit-unsafe             40     1     41   (3%)
  irrational-onset        23     2     25   (2%)
  no-leaf-anchor           2    14     16   (1%)
  view-unusable            4     6     10   (1%)
  resolution               0     6      6   (0%)
  mixed-pitch-domain       0     2      2   (0%)
```

This was previously obtainable only by patching the real functions with a temporary recorder and reverting. It is now a committed test.

## About the percentages — a correction worth stating plainly

Going in, the expectation was that separating "wrong surface" from "refused" would move our editability numbers upward with no behaviour change, and that this should be read as a reporting correction rather than progress.

Measured, it is narrower than that. `wrong-surface` is 67% **of this sweep**, which asks BOTH surfaces of every unit — so a drum pattern declining the piano roll is counted, and it should be. That is where the inflation lives, and every figure quoted off a both-surfaces sweep needs reading with it in mind.

The coverage harness never had that problem: it routes by head (`note`/`n` → roll, `s`/`sound` → grid) exactly as the app does. Over the vendored corpus its `wrong-surface` count is **zero**, and its headline percentages do not move at all. Under head routing the gate keeps its teeth — a `wrong-surface` unit is one whose head asks for a view its own values cannot fill, and the user gets code.

So: same gate, two honest readings, and no free percentage points. The blocker histogram is now keyed by gate, which is the part that changes what we can see.

## Also folded in

"Anchors are valid only for the layout they were read against" was enforced independently in both leaf writers — the grid via `cols.length`, the roll via `leafSource.steps` (added in #989 after it was found splicing `~` over notes during a resize). One invariant in two places is how the next surface ships without it; both now call `anchorsDescribe`.

## Verification — this phase must move no verdict

- **0 verdict flips** over all 1500 corpus units × both surfaces, sweeping the shipped parsers against the pre-change ones side by side; 1556 reason-only moves.
- The updated snapshots **add and remove exactly zero `"ok"` lines** (counted programmatically, not by eye).
- **writer-reach floors hold at 95 / 73**, zero losses — the report now also breaks unopened units down by gate, which is what #991 needs to size the period cap honestly.
- New `refusal-gates.test.ts` (6): every projection refusal carries a gate · the core's feature vocabulary never surfaces on a gated refusal · the gates are non-vacuous · surface mismatch is separable · the former "nested groups" units name a real gate.
- editor 2985 · app 911 · parity-corpus 305 · editor tsc 62 baseline, 0 new · app tsc 1 pre-existing · eslint no new findings.
- **Observed in the real app** through the rebuilt dist: 23 e2e green (`sequencer`, `sequencer-projection`, `pattern-grid-coverage`) — projected grids, nested groups, bar expansion, and the standby fallback all behave as before.

`EDIT-COVERAGE.{md,json}` are deliberately not staged: they carry an uncommitted regeneration from the #986 reach work, and bundling it here would make a legibility change look like a reach gain.

## Deliberately not in scope

Unifying the syntactic writer onto `serializeByLeaf`, and deleting the ~25 syntactic guards in favour of the invariant itself. Both belong to the bijection statement that follows #991.


MrityunjayBhardwaj:
## Self-review — three passes, two findings that mattered

Two follow-up commits on top of the original.

**Pass 1** (`198180f1`) — three legibility findings, all mine:
- The `mixed-pitch-domain` gate reused the core's own "…is beyond the editable subset" phrasing, which forced the new vocabulary test to carry an exemption. Reworded; the test is now strict with no exemptions.
- `writer-reach`'s new tally guarded on `m === null` and then had to spell a branch that can never be taken. Narrowed on `!r.ok` directly.
- `element-tiling` is never surfaced (refusals report the leaf writer's gate) with nothing saying so. Now says so, and says why it is kept rather than deleted into something vaguer.
- `numericNotes.test.ts` asserted on the wording; it now asserts the **gate**, so re-phrasing a sentence cannot turn a discriminator red.

**Pass 2** (`dba646d9`) — a real one, and it lands on the number this sequence was planned from.

`leafAnchors`/`rollAnchors` returned a bare `null` for two unrelated things: a note with no usable source span (the write-back guard) and a note that does not fit inside the bar it starts in. Both were reported as `no-leaf-anchor` — the same misattribution the whole phase exists to end, one level down.

Measured, 6 of the 16 were the second kind. Those notes have a perfectly good source token; the column layout is what refuses them.

```
  no-leaf-anchor          2     8     10  (0.6%)   ← the write-back guard
  note-crosses-bar        0     6      6  (0.4%)   ← was folded into the above
```

So the write-back guard is **10 units, 0.6%** — smaller still than the 1% the #990 → #991 ordering was decided on, and in the same direction. It does not change the sequencing; it does mean the residual table in the issue should be read with this split.

Neither pass moved a verdict: floors hold 95/73, and each snapshot move is exactly the reworded reasons (0 `"ok"` lines added, 0 removed, counted programmatically). e2e re-observed green after each rebuild.

PR #1004, opened 2026-07-24
🔢 fix(harness): ask every shipped editor before calling a unit uneditable

MrityunjayBhardwaj:
Closes #1003.

## Problem

`editCoverage.ts` decides whether a unit has a view by consulting a hand-written set of detectors: `parseStepGrid`, `parsePianoRoll`, `detectAllArrangeCalls`, `detectAllPickControls`. `detectMasterAll` (`mixer/masterEdit.ts:77`) shipped after that set was written, and nothing tied the two together.

So every master `all(x => …)` line has been measured as `code-only` — **16 units across 150 real Bakery tunes that the mixer's master strip can edit today**:

```
all(x=>x.gain(1.5))
all(x=>x.clip(1).transpose("<0 3 6 6 3 0>/16"))
all(x => x._pianoroll({vertical: 19}))
```

This is the mirror of #998. There the *denominator* held statements no view could ever serve; here the *numerator* misses units a shipped view already serves. Both come from the same place: a definition written once and never re-derived from what actually shipped. And `code-only` is read as "no view exists for this yet" — a roadmap item — when all it ever meant was "none of the detectors this file happens to import said yes."

## Fix

**Consult it, and report it as its own class.** `master` sits beside `note` and `clip` rather than folding into either, so the surface stays visible in the report instead of silently inflating another bucket — the shape #998 settled on. Counted as structurally editable: a fader, pan and insert chain is real view editing, not a bare numeric argument.

**Then stop it recurring.** `detector-coverage.test.ts` scans `visualEdit/` for exported `detect*` entry points and fails when one is neither consulted nor listed in `NOT_CONSULTED` with a reason. The exclusion list is the load-bearing half — an entry there is a claim that consulting the detector would make the measurement *wrong*, not that nobody has got round to it. All four current entries say which: three are cursor-position or subset variants of detectors already consulted, and `detectBarePattern` returns a wrap target for any bare pattern rather than a view verdict, so consulting it would classify the whole corpus as editable and measure nothing.

## Test plan

- **Both directions red-tested.** Renaming the import away → fails naming `detectMasterAll`. A new `export function detectProbeSurface` appearing in `visualEdit/` → fails naming it. Green again after reverting each.
- **The first version of the gate could not fail.** It used `harness.includes(name)`, which still matches when the import is renamed to `detectMasterAllXX`. Caught by red-testing, not by reading it. Now a word-boundary match, with the reason recorded at the line.
- **Verified against the real consumer, not inferred:** `useMixerModel.ts:164` binds the master line and `MixerStrips.tsx` renders its gain/pan/mute edits, so this is a wired surface rather than a detector that merely exists.
- **The one over-count this could have had, measured:** `detectMasterAll` returns every master line while the mixer binds one. Across the corpus and the 150-tune sample: 16 master lines, all 16 the bound line, 0 unbound extras, 0 mute sentinels, 0 viz-only. Recorded as a stated assumption at the classification site rather than a gate for a case that does not occur.
- **Snapshot drift is one line and explained.** `bakery-158-all-shape-fence` moves knobs-only → partial (2 units, 1 now editable). Its fence is IR parity, not editability, and its parity verdict is unchanged. Never `-u`'d blind.
- parity-corpus **334 ✓ / 19 files** · full app suite **940 ✓ / 64 files** · `tsc` 1 error, pre-existing and identical on the base. No editor source touched, so no `dist` rebuild.

## Effect

Corpus: structural 54 → 55 of 101 (53.5% → 54.5%), `code-only` 12 → 11. Numerator only, one-way — nothing moves out of editable.

Over the 150-tune sample the same change is **+16 units, 547 → 563, 61.5% → 63.3%**.

This unblocks the invariant-3 restatement discussed in #1001 and #1002: a restatement calibrated against a numerator that under-reports would be calibrated against the wrong number.


MrityunjayBhardwaj:
## Critical self-review

Two findings, both fixed in `f0df81de`:

**1. The master check could double-count.** `collectUnits` expands stack/cat voices anywhere in the document, so voices inside `all(x => x.add(stack(a, b)))` lexically overlap the master statement — and the strip edits the whole LINE, not any one voice. Crediting both would inflate the numerator, the same over-count this PR exists to stop, pointing the other way. Guarded on `!u.nested`.

**2. Then I measured the guard and it fires zero times.** `detectChunk` at those positions anchors on the enclosing statement and the range-dedup collapses them, so no distinct nested unit survives to be miscredited. Corpus numbers are byte-identical with and without it. The comment now says DEFENSIVE and why it is kept regardless — a guard that reads as load-bearing when it never fires is a false claim about the code, and this repo has been bitten by the inverse (a passing test read as evidence a guard was working).

**A third, in the test I wrote for it:** the first draft asserted two note voices inside the master line and failed. `x.stack(…)` has a member callee, so it is never expanded — the shape I was guarding against is not the shape I wrote. The assertion now records observed behaviour instead of a prediction.

Nothing else in the diff is load-bearing beyond what the description covers: the remaining changes are report plumbing (`master` threaded through `TuneReport`/`aggregate`/markdown/JSON) and they are exercised by the regenerated report.

Re-run on the final tree: parity-corpus **335 ✓ / 19 files** · `tsc` 1 error, pre-existing and identical on the base.

PR #1015, opened 2026-07-25
🧭 Eval-first resolution of a unit's note-content span (#1006 P0, epic #1007 P1)

MrityunjayBhardwaj:
Phase **P0 of #1006**, which is **phase P1 of epic #1007**. Ships the resolver **inert** — nothing routes through it yet.

## What was wrong

Deciding *which document span carries a unit's note content* was a parse-side rule that only ever saw the head call's first string argument. Four shapes escaped it and were each filed as their own feature — a chained argument (#932), a bound reference (#931), a root literal, and head routing — while the engine had been answering all four for free: every mini atom is created with its document span attached, so the span rides the hap.

Widening the parse rule **by name** was tried and refuted: it resolved `introFanfare.s("gm_trumpet")` to the trumpet, 30 times over, and the parse gate, the writer gate and the engine round-trip all passed it — deleting the only cell of a one-cell grid *is* a faithful edit. `s("bd sd")` as a head is content; `.s("piano")` after a pattern source is timbre. Only position can tell them apart.

Two further faults made evaluation look like the wrong answer, and both were diagnosed before this branch existed:

- **#1014** — a hap location is document-space *only if the transpiler rewrote that string*. Runtime-synthesised strings (`voicing`) and single-quoted literals return **correct offsets in a coordinate space nobody declared**, into the same flat array. `note('c4 e4')` returns two bogus spans and no correct one, while sounding perfectly.
- **#1013** — a location's *presence* is a function of the **query window**, which scales with the pattern's **period**. A four-cycle window loses 45% of one quoting class, and because long patterns are the ones written with backticks, that reads as a syntax bug.

## What this does

**Eval proposes, the AST disposes — and neither half does the other's job.**

| | |
|---|---|
| **Admission** | Ask the transpiler, don't filter: only spans present in `meta.miniLocations` are candidates. Removes the coordinate-space leaks *by construction* (86 across the corpus). A structural filter cannot do this — `const` is not syntactically anybody's argument. |
| **Disposal** | One bit per candidate: pattern **source**, or control **argument**? Decided by POSITION, never a vocabulary of head names. Follows bindings, so `.bank(drums)` disposes as an argument exactly as `.bank("Linn9000")` does. Bounded at the unit, so `"<0 1>".pick([intro, core])` consuming whole voices as arguments does not turn all of their contents into control values. |
| **Window** | `QUERY_CYCLES` and the reason it is what it is live in the contract, not in a probe. |
| **Fallback** | The parse walk is the *same disposal with a different proposer*, and serves **per unit** whenever evaluation is silent. |

The one name-based exception is `note`/`n` mid-chain, and only on a chain with no head call of its own — `"gm_pad_warm".note("<c4 e4>")` overrides a root literal, `sound("hh").note("c2")` does not override the `sound(…)` the harness edits. It is stated, isolated, and red-tested.

## Recalibration came first

#1006's stated gate (presence 81.8% / filter-exact 77.7%) came from the instrument later found to be under-windowed and unmirrored, so it was re-derived rather than built against. The harness was calibrated against a known answer before anything else was read: it reproduces the committed baseline — **889 units, 563 structural** — exactly.

Raising harness eval coverage is #1008's job, but the calibration denominator needed it honest, so span-neutral shims for the editor-side visual taps landed here: **60.7% → 88.7%**. Those shims and their reasoning are #1008's starting point.

## Measured — 150 real tunes, offsets 0/250/500

```
eval coverage      133/150 (88.7%)
undeclared spans   86        (coordinate-space leaks, refused by admission)
known-content      485
  content present  450/485 (92.8%)
  resolved exactly 485/485 (100.0%)   wrong 0   refused 0
  of those exact, 35 answered by the parse fallback
unoffered pool     148/173 (85.5%) got a SPAN — not a view, not an edit
```

**The 35 misses are fully diagnosed, and it is the sharpest thing here.** Every unit whose content span is absent from the located haps belongs to a statement with **no located hap at all** — a bare pattern statement that is not the document's last never sounds, because the transpiler returns only the last statement, and when `$:` collected anything the repl replaces the return with the stack. Not one unit is a case of "the statement sounds but this mini has no span", and the calibration **asserts** that class is empty. That is the answer to the 24 unexplained double-quoted misses carried into this phase.

**Do not read 148/173 as reach.** It is a span and nothing more; opening a view, the view having structure, and a modelled edit surviving the engine are three further gates, and the last measurement of that chain converted 173 unoffered units to 97 → 65 → **47**. A one-cell view of an instrument name is a *correct* resolution and a *useless* surface.

## How this was checked, given 100% is a suspicious number

The calibration population is units the harness already calls `note`, where the truth and the disposal's head-call arm are close to the same statement — so agreement there is a **regression fence, not evidence**. It says so in the file. The evidence is elsewhere:

- **Red-tested.** Each of the four load-bearing properties was broken in turn and the failure required: emptying the note override (1 test), flipping the mid-chain arm to `source` (3), removing the unit boundary (1), flipping the override arm (2).
- **The ranking can hide a broken disposal** when the content literal is written first, which it usually is — so the control-argument cases assert on `alternatives` being empty too. That is the load-bearing half of that test.
- **Hand-read the pool where the rule can be wrong.** All 173 unoffered resolutions were dumped and read. Every suspicious one-token result (`s("o1")`, `s("breaks165")`, `sound("sawtooth")`) is a genuine one-token pattern source; `n(run(13)).sound("cowbell")` is correctly **refused**.

## Test plan

- `pnpm --filter @stave/editor test` — 207 files / 3021 ✓, tsc unchanged at its 62 pre-existing errors
- `pnpm --filter @stave/app test` — 65 files / 942 ✓ (941 before, +1 = the calibration)
- `tests/parity-corpus` — 20 files / 336 ✓
- editor `dist` rebuilt and byte-identical, which is what "inert" means here: the module is not reachable from the package barrel

`unitsWithStatus` is extracted from `measureDocs` so the calibration can ask the harness per unit instead of keeping a second copy of the unit model beside it; `measureDocs` now calls it, verdict-neutral and verified by the committed baseline.

Closes #1014. Part of #1006, #1007, #925. Supersedes the standalone framing of #931 and #932.


MrityunjayBhardwaj:
## Critical self-review — three gaps found and fixed (`e0fdc3cb`)

100% exact on the calibration is what the *calibration* says about a population where the rule can barely be wrong. Auditing the diff directly rather than trusting that number turned up three real defects.

**1. Attribution walked every identifier that SPELLED a binding's name, not its references.** Method names are identifiers too, and Strudel's vocabulary collides with the names people give their patterns constantly — `.cpm(…)` beside `let cpm`, and `.p1`/`.d1` (the repl's own pattern getters) beside `let p1`/`let d1`, which is a real document in the corpus. That pulled unrelated initialisers into a unit's reachable set, making **another statement's span admissible as this unit's content**. Measured: **37 of 148 documents** carry at least one such collision. Fixed by walking `refs`, which already excludes member properties and object keys.

**2. The root-position override required the note argument to be a LITERAL.** So `"gm_pad_warm".note(mel)` left *both* strings disposing as `source`, and the answer came from whichever was written first:

```
$: "gm_pad_warm".note(mel)   /  const mel = "<c4 e4>"   →  gm_pad_warm   ← wrong
const mel = "<c4 e4>"        /  $: "gm_pad_warm".note(mel) →  <c4 e4>    ← right
```

Same program, two orders, two answers. Any argument counts now. A bare `.note()` with **no** argument still does not override — that is the opposite operation, reifying the root itself as note content (`"0 5 3 2".scale('G4 minor').note()`) — and both directions are pinned.

**3. A comment claimed something the code never checked.** It said a binding used both as a pattern and as a control argument was "reported by the calibration rather than hidden". The calibration did not report it. It is **11 of 192 bindings**, and it is now printed.

### The ranking was hiding two of these

Both (2) and the earlier control-argument cases pass on `text` alone whenever the content literal happens to be written first — which it usually is. `alternatives` is where a wrongly-admitted span shows up, so the affected tests now assert on it too. That is the load-bearing half of those assertions.

### Verdict-neutral, predicted before the run

eval coverage 133/150, 485/485 exact, 148/173 unoffered resolved, mixed-use 11 — all unchanged. So (1) was **latent** in this corpus rather than live, which is exactly why it needed a red-test and not a measurement to catch.

### The gate now fences six properties

Each broken in turn and required to fail — the note-override set, the mid-chain `argument` arm, the override arm, the unit boundary, the reference-based attribution, **and a revert to the previous version of the override rule**:

```
RED note override emptied              → 2 failed
RED mid-chain arm returns source       → 3 failed
RED unit boundary removed              → 1 failed
RED override fires unconditionally     → 1 failed
RED override requires a literal (prev) → 1 failed
RED attribution widened to all idents  → 1 failed
RESTORED                               → 20 passed
```

Gates: editor 207 files / 3023 ✓, tsc unchanged at 62 pre-existing · app 65 files / 942 ✓ · editor `dist` rebuilt, byte-identical.

### Known and deliberately not fixed here

- **`QUERY_CYCLES = 128` is a MEASUREMENT window.** 128 `queryArc` calls per document is right for a corpus sweep and wrong for a live editor, which has the running pattern's haps already. P2 (routing the app) has to take them from the engine rather than re-query.
- **`evalHarness` mutates `Pattern.prototype` process-wide** (`.p`/`.q`, the visual taps). It is test support with one importer; a second importer in the same worker would inherit the shims.
- **The mixed-use tie-break stays "source"** — losing real content is the worse error — but it is a judgement call over 11 real bindings, not a proof.

Issue #1010, opened 2026-07-24
✂️ P4: retire syntax authoring down to the irreducible — surgery wherever a span exists


MrityunjayBhardwaj:
### Why

`notation/serialize.ts` holds two families that do opposite things:

- **`reemit*`** — `reemitRegion`, `reemitStep`, `reemitAltRegion`, `reemitRollRegion`, `reemitAltRoll`, `groupWrapRegion`, `laneString`, `gridBars`, `rollBars`, `packLanes`, `toPlaced`, `assignNotes`, `buildGroups`. These **author Strudel syntax from a model.** That is modelling: we decide how the pattern should be spelled.
- **`serializeByLeaf` / `spliceByLeaf` / `spliceRollByLeaf`** — byte surgery at the note's own span. That is the **adapter**: unedited bytes are copied verbatim and the engine's own spelling survives.

The adapter already exists and already carries both surfaces. This phase finishes routing to it.

### What to do

Route every edit that has a span through the leaf writer. What still calls `reemit*` afterwards is, by construction, an edit with **no span to splice** — and that set should reduce to the epic's stated irreducible:

1. adding structure where the source has no leaf (a resolution change — subdividing, not replacing an existing `~`)
2. the signal tail, which has no source span at all
3. `@n` duration, which is never a location and is refused by principle

**Amended 2026-08-18, after P4e measured the residue instead of reading it (shipped in #1296): it is FOUR, not three.**

4. a delete whose leaf is **shared between columns** (#1160), or whose unit carries **no leaf index at all**. A delete *has* a span — the note's own bytes — and surgery answers most of them; the remainder fall through and author notation anyway, for an edit that had bytes to replace. **This case is not irreducible**, which is why it is tracked as #1295 rather than accepted here.

And two corrections to the three above, from the same measurement, so this list stops describing something the code no longer does:

- **Case 1's parenthetical is the half that does not land.** "A resolution change — subdividing" reaches no surviving `reemit*` site at all; the part rebuild inside the splice path answers it. Case 1 rests entirely on its other half — placement into a slot the source never indexed.
- **Case 2 reaches no site either.** The gain writers call nothing in this family, so the signal tail was never one of these writers' populations.

Per-site figures, the population and the recipe live in the map block at the top of `notation/serialize.ts`. They are a dated observation, not a live gate — re-take them before quoting the digits as current.

### Done when

[...]

PR #1033, opened 2026-07-26
📏 fix(notation): the grid reader keeps each note's length — P4's first step, deliberately inert

MrityunjayBhardwaj:
First two commits of P4 (#1010). Both land alone and on purpose: the reader change is the **control arm** for everything that follows, so it must move no number, and it doesn't.

## Why this comes before the routing change

I started P4 as the issue frames it — route edits through the leaf adapter and comment whatever still needs the printer. Measured, that flip is a real win on the grid (+5 reach, 16 fewer silent length rewrites) and a real loss on the roll (−10). But it leaves 13 units still opening via the element writer and still silently changing a note's length, and the only choices for them are "accept the corruption" or "take editability away from 13 views."

That trade is manufactured by a lossy model, not by the routing:

```
PianoRollModel   notes: { pitch, start, duration, gain }   ← a note list
StepGridModel    lanes: [{ sound, cells: boolean[] }]      ← a bit matrix
```

The roll has never produced a duration loss. Same writer families, same element-first routing — the difference is that its model can hold what the writer needs to preserve. `readGridOnsets` made this literal: it typed the hap as `whole?: { begin: … }`, with `end` absent, so the grid could not read the axis even in principle.

So the routing lands later, as a consequence of the model being right, rather than as the fix. Plan in full on #1010; the structural generator behind it is #1032.

## What's here

**`Onset` carries `durs`**, index-aligned with `atoms` and `spans` — the parallel-array idiom `spans` already establishes — read exactly the way the roll reads it (`whole.end - whole.begin`) so the two surfaces cannot disagree about what a note's length is.

**The roll's routing comment is corrected.** It recorded leaf-first as costing the roll one unit (73 → 72), measured at #994 on an onsets-only oracle. Re-measured against the duration-aware oracle it costs **ten**: 75 → 65. Not wrong when written — taken with an instrument that could not see the axis the flip moves, so it never transferred. It has been sitting in the file as a reason to trust a decision at ten times the confidence the evidence supports.

## Test plan

**Inert, as designed** — writer-reach **126** step / **75** roll, projected **186** / **118**, parity-corpus **346/346**, tsc **62** (baseline). All unmoved.

The zero is meaningful because the detector is proven to move: the two routing counterfactuals measured on this branch shifted these same numbers to 131 and to 65, then were reverted and the baseline re-measured to confirm the instrument had not drifted.

**Observed, not inferred.** "Unmoved" is equally consistent with `durs` being correctly populated and with it being silently empty, so the field is checked directly against known unequal lengths:

| pattern | reads | |
|---|---|---|
| `bd@3 sd` | 0.750, 0.250 | the `@3` weight, not the column width |
| `[hh ~]!16` | 16 × 1/32 | on a 1/16 grid — the corruption case, now visible |
| `[bd@0.5 - - -]` | 1/7 | 0.5 against 3.5 sibling weights |

`gridOnsetDuration.test.ts` pins all five cases and **fails all five** when the axis is broken (verified by breaking it). Editor suite 211 files / 3045 tests.

`readGridOnsets` is exported for that test alone, documented at the export. Deliberate: the boundary that drops axes is the one worth being able to interrogate, and a field nothing asserts is a field that drifts.

Refs #1010, #1032

---

### ⚠ Figures re-based later in this stack

The corpus harvester was rebuilt in #1043 — it stopped approximating the transpiler with a
regex and started asking it, so `mini-corpus.json` went **1500 → 1535** units (backtick minis
in, commented-out code out). That moved every reach figure this project quotes.

**The numbers above were measured on this diff's own tree and are correct there.** Re-verified
per commit across the whole stack: each commit's own harness against its own tree, and the only
movement anywhere in the seven is at #1043. No writer changed across that move — the population
did.

Superseded at the stack tip: writer-reach **126 step / 75 roll → 123 / 85**.


MrityunjayBhardwaj:
### Self-review

Two findings in this diff, both from probing the change rather than re-reading it.

**#1034 — the dedupe drops a second length at one instant.** The new `durs.push` sits inside the existing `atoms.includes(token)` guard. Atoms genuinely collapse per column; duration does not, and the guard treats them alike. `bd*2, bd` records `durs=[0.500]` and drops part B's 1.0. Worse than lossy — *which* length survives depends on hap arrival order, which is not uniform, so it is unstable. Nil impact here (nothing reads `durs` yet, which is why this PR is inert) and load-bearing in P4b, so it is filed rather than carried forward.

**Placeholder `0` durations.** The synthetic probe onsets in `leafEditSafe`/`leafExpected` carry `durs` of `0` and `[]`. Only `onsetKey` reads them and it ignores duration, so they are correct today, but `0` is a magic value meaning "unknown" and would read as a zero-length note to anything added later. Noted in #1034 rather than fixed here, to keep this commit inert.

**My first probe of the dedupe was wrong** and worth recording: I tested `[bd@2, bd]` and `bd,bd@2`, which cannot exhibit the bug — a `,`-stack normalizes each part to the full cycle, so every reading was 1.0 and the gap looked absent. The case needs two parts that disagree *within* a cycle (`bd*2, bd`). A probe that cannot produce the failure is not evidence of its absence.

### Gates, all packages

editor **3045/3045** (211 files) · app non-corpus **606/606** · parity-corpus **346/346** · tsc **62 editor / 1 app**, both at baseline · writer-reach **126 / 75**, projected **186 / 118** — unmoved, as intended.

PR #1035, opened 2026-07-26
🧩 fix(notation): the grid reader keeps every note at a column, not the first of each name

MrityunjayBhardwaj:
Stacked on #1033 (base `fix/1010-retire-syntax-authoring`), landing alone and additive per the plan on #1010. Closes #1034.

## The defect

`readGridOnsets` deduped a column on **atom identity** and filled the span and the length inside that same guard. The predicate is about what a cell *displays* — right, a grid cell shows a sound once — and it was also deciding what the reader *retains*:

```
"bd*2, sd"   atoms=["bd","sd"]  spans=[{0,2},{6,8}]  durs=[0.5,1.0]
"bd*2, bd"   atoms=["bd"]       spans=[{0,2}]        durs=[0.5]     <- part B gone
```

Same stack, same two haps; the only difference is whether the two tokens happen to be spelled alike. Where a `,`-stack put one sound at one instant twice, the second note was dropped whole — length *and* source anchor — and which survived came down to hap arrival order.

## Measured before choosing

Sweep report: `packages/app/tests/parity-corpus/SWEEP-1034.md`, arms `_sweep-1034{,b,c,e,f}.spec.ts` (`_`-prefixed specs sit outside the gate's include glob).

| population | denominator | units affected |
| --- | --- | --- |
| 150 tunes, offsets 0/250/500 | **889 musical units** | **2** |
| `mini-corpus.json` | 1500 distinct minis / 360 tunes | **4** |

Two populations, never summed. The doc arm reproduces 889 exactly from the imported unit model, which is what says it is the right denominator rather than a lookalike.

Every instance is a `,`-stack sibling; **zero** rounding collisions, the other mechanism worth fearing, which was looked for specifically. Cost of fixing: **zero writer-reach** — three of the four never project, and the fourth is owned by the syntactic core, which writer-reach excludes by construction.

## Duration was the smaller half

Classified per dropped occurrence over 4662 unit×cycle pairs:

| what the dropped occurrence carried | occurrences | units |
| --- | --- | --- |
| a different LENGTH — the filed defect | 8 | 4 |
| a different ANCHOR (leaf span) | **32** | **9** |
| identical on both axes | 12 | — |
| **total dropped** | **44** | **10** |

The anchor is the write-back target, so those are edits landing on one of two source leaves and silently ignoring the other. `[b4,d4,f#4],b4@4` makes the point: both haps last exactly 1.0, so **no length is lost and a duration-shaped fix would never have seen it**, while the anchors differ (`{1,3}` vs `{12,14}`). That axis had been collapsing since before the length was read at all.

## The fix, and why this shape

The column records every hap as `Occurrence { token, span, dur }`; the dedupe moves to `deriveColumn`, which builds the display view. `atoms`/`spans`/`durs` stay, derived by literally the loop that used to fill them in place.

Parallel arrays make "stay index-aligned" the invariant the code advertises, so adding a fourth `push` beside the others *preserves* it while importing a collapse rule only ever justified for the first — which is how the length inherited the anchor's bug a session after it was written, with nobody noticing either time. One array of records cannot go out of alignment and leaves no guard for the next axis to be added inside; gain (#1027) and the unread `part` (#1028) become fields here.

Refusal stays with the writer. With both occurrences recorded the leaf adapter can write to the correct leaf because it now knows there are two; only the element re-emit, which must re-derive, needs to decline on disagreement — decided on complete information instead of guessed from arrival order.

Synthetic probe onsets now carry `dur: null` meaning NOT KNOWN, replacing a placeholder `0` that a later consumer would read as a zero-length note.

## Test plan

- **A/B against the actual pre-change reader** (`git show <base>:parse.ts` imported beside the new one) over the corpus: **4662 pairs compared, 0 differences** in `atoms`/`spans`/`durs`; 19375 occurrences retained against 19331 atoms after dedupe. Inertness proven by observation rather than by construction.
- **Five new cases** in `gridOnsetDuration.test.ts`, written from the sweep's independently observed values. **Four go red** when the guard is restored; the distinct-token case stays green, so the change is scoped to same-name collisions.
- Figures read off the report, not inferred from the floors passing — those are `>=` assertions and a rise would pass silently: writer-reach **126 step / 75 roll**, projected **186/118**, losses **29 (leaf 0 / element 29)**.
- parity-corpus **346/346** · editor **3050** (3045 + 5 new) · app non-corpus **606/606** · tsc **62 editor / 1 app** (both baseline).
- `pnpm --filter @stave/editor build` run and the whole `dist` committed.

Every number that existed before this branch is unmoved, which is the point of landing it this way.

---

### ⚠ Figures re-based later in this stack

The corpus harvester was rebuilt in #1043 — it stopped approximating the transpiler with a
regex and started asking it, so `mini-corpus.json` went **1500 → 1535** units (backtick minis
in, commented-out code out). That moved every reach figure this project quotes.

**The numbers above were measured on this diff's own tree and are correct there.** Re-verified
per commit across the whole stack: each commit's own harness against its own tree, and the only
movement anywhere in the seven is at #1043. No writer changed across that move — the population
did.

Superseded at the stack tip: writer-reach **126 / 75 → 123 / 85** · losses **29 → 30**
(still leaf 0 / element 30) · corpus **1500 → 1535**.


MrityunjayBhardwaj:
### ⚠ Correction to the figures above — they carried an unstated cycle window

The sweep that produced the counts in this thread read **cycles 0-3**, and nothing derived that 4 from anything. Auditing the measurement apparatus end to end (#1039) found the window is itself a restriction, and a load-bearing one:

| window | accepted pairs | notes kept | collision units | dropped occurrences | duration-disagreeing units |
| --- | --- | --- | --- | --- | --- |
| 4 | 4662 | 19375 | 10 | 44 | 4 |
| 8 | 9328 | 38770 | 10 | 87 | 4 |
| **16** | 18662 | 77659 | **11** | **182** | **5** |
| 32 | 37326 | 155512 | 11 | 364 | 5 |

Converges at 16. The unit that requires it is

```
bd*2,[- sd]*2,[- hh]*4, <-!7 oh>, <-!12 bd*4 bd*8 bd*16!2>
```

whose alternation does not reach its colliding arm until **cycle 12**.

**So:** the mini-corpus counts stated as *4 units / 8 instances* are **5 units / 11 with any same-token collision**. The axis split stated as *8 duration / 32 anchor / 12 duplicate across 10 units* is **35 / 134 / 48 across 11 units**.

**What does not change:**

- the doc-population figure — still **2 of 889**, so that one is window-robust;
- the ratio — anchor still roughly four times duration, at every width measured, which is what the decision to fix the representation rather than the axis actually rested on;
- the cost — the fifth unit is `unstable-period` on the grid and `wrong-surface` on the roll, so **writer-reach at stake is still zero** (verified, not assumed);
- every product figure — 126 / 75, 186 / 118, losses 29.

The conservation gate now runs at 16 cycles and pins the window in its assertions next to the population, so this particular restriction cannot go unstated again. Noting it here rather than editing the comment above, because the correction is the useful part: this is the fifth figure produced at this boundary carrying a restriction nobody wrote down, and I committed it in the same session as the write-up naming the other four.

Issue #607, opened 2026-06-27
Pattern tab: slot-count robustness + velocity consistency across slot changes


MrityunjayBhardwaj:
Investigate the EXPECTED behaviour first (Logic Pro + the reference player) for how controls/velocity behave as the slot count changes, then align our Pattern tab. Velocity must stay consistent across slot adjustments.

MrityunjayBhardwaj:
Partially addressed in #626 (merged to studio @ 21c41ea5): **single-bar** slot increase is now conservative — adding slots keeps each note a single slot (held `@n` preserved), start maps proportionally, and per-note velocities carry through + re-align (new empty slots serialize as `~`); a note placed in a new slot is immediately velocity-editable. Leaving OPEN for the deferred bits: **multi-bar `<…>`** still uses the lossless ×2 path, and the optional drag-on-empty-slot-creates-note-with-velocity gesture.

MrityunjayBhardwaj:
Landed via PR #627 → studio_v0.2.0 (merge 47a5244c). The conservative multi-bar slot increase shipped; the empty-slot drag-to-create velocity gesture was dropped after live testing (an empty velocity slot has no note, so a velocity gesture there reads as spurious note creation — empty slots are now inert). Both of #607's deferrals are resolved (multi-bar landed; empty-slot abandoned by design), so closing as done.

PR #1044, opened 2026-07-26
📐 P4b — the step cell carries how long its note sounds

MrityunjayBhardwaj:
Seventh in the stack: `#1033` → `#1035` → `#1039` → `#1040` → `#1042` → `#1043` → **this** → `studio_v0.2.0`. Based on `fix/1037-harvest-quote-styles`, so please read that one first.

⚠ **Every figure in this PR is on the rebuilt 1535-unit corpus.** The PRs below this one carry figures that `#1043` then re-based; nothing here quotes a pre-rebuild number.

## Why

A step cell was one bit. So how long its note sounds was not in the model — and a writer cannot preserve an axis its model never carried. That is where every duration loss on this surface starts: the element re-emit has nothing to write with except the view's own resolution, so `[hh ~]!16` — sixteen notes each sounding for **half** a column — comes back as sixteen notes of a full column, twice their length. The piano roll's note has carried a duration since the beginning and has produced zero of these.

## What changed

`StepLane.cells` becomes `StepCell[]`: `false`, or a note with a `duration` **in columns**.

- **`false`, not null/undefined** — every reader that only asks "is anything here?" keeps working unchanged (truthiness, `.some(Boolean)`, `filter(Boolean)`).
- **An object, not a bare number** — a `0` would read as OFF, and gain (#1027) / `part` (#1028) become fields here rather than another shape change (#1032's direction).
- **Columns, not cycles** — `Onset.durs` is cycle-relative so no reader needs the grid's resolution; a cell is already in the grid, and every consumer reasons in columns. `RollNote.duration` is the same unit. Fractional is normal: ~206 of 1535 units carry a length that is not 1, and 5 are sub-column.

All three read paths populate it — the syntactic core from the slot's span, the projected paths from `whole.end - whole.begin` scaled by columns-per-cycle, the leaf path from the anchors paired with the played lengths.

Each op's rule follows the promise that op already made:

| op | rule | why |
|---|---|---|
| ×2 / ÷2 | scale | "every hit keeps its position", now covering the whole note |
| resize spread | scale | preserves musical time |
| resize pad | keep | preserves step indices, so the groove stretches |
| quantize | scale down when coarsening, keep the column count when refining (#607), merge to the shorter, clamp to the next hit | mirrors `quantizePianoRollTo` |
| a painted hit | one column | the box the user clicked |

÷2 needs no integrality guard, and that is worth stating because the roll needs one: `RollNote.duration` counts whole `@n` steps, so an odd length cannot halve and `canHalvePianoRoll` refuses. A cell's length is fractional, so ÷2 represents exactly, can only ever **shorten** in column terms, and drops nothing — `canHalveStepGrid` is unchanged.

## Two defects the new gate caught on its first runs

Both are in this diff, not pre-existing:

- a `,`-stack part coarser than the shared grid kept **its own** column count as the length. `sd` beside `bd bd` is one note over a two-column grid: two columns, not one. **69 cells over 14 units.**
- quantize merged *before* scaling, so a value already scaled by an earlier source cell in the same bucket got scaled twice — two 1-column notes merged to 0.25 instead of 0.5.

## Test plan — a representation change that moves a number is not one

- **writer-reach 123 step (floor 123) · 85 roll (floor 85) · losses 30 (leaf 0 / element 30)** — all unmoved
- parity-corpus **349 → 352** (the three new gate arms) · editor **3050 → 3063** (thirteen new cases) · app non-corpus **606** · editor `tsc` **62 errors, the same 62**
- **A/B of every grid op against the pre-change code over the whole corpus: 19360 invocations, 0 differences in ON/OFF.** Shown to fire by breaking ÷2 to keep the odd columns (10 differences), then restored.
- **the field observed directly**, because unmoved numbers are equally consistent with an empty field: lengths unequal within one pattern (`bd hh*2 sd cp` → 2 and 1), longer than a column (`bd [sd sd sd]` → 3), shorter (`[hh ~]!16` → 0.5), neither (`[bd@0.5 - - -]` → 1/7). **Red-tested** — with the axis broken, 7 of 13 cases fail and the corpus gate reports 816 mismatches instead of 0.
- **browser gate: 53/53** (`pnpm gate:editing:browser`) — real cell clicks, velocity drags, round-trips.

## The new gate

`cell-duration.test.ts` asserts every ON cell's length is a length the **engine** actually played — **4763 cells across 968 units** (783 syntactic / 113 derived / 72 leaf), over each model's own cycle window, population pinned, with a control arm proving a degenerate reader is caught. "The slot's span is the note's length" was an inference from both being integral; it is now measured.

Its sampling window is each model's **`bars` cycles**, not a fixed 4 or 16: a grid's cells cover exactly `bars` cycles, so a hap at cycle 5 of a single-bar model is a hap no cell exists for. The first cut of the sweep sampled wider and talked itself into a defect that was not there — that write-up is in the first commit.

## Commits

1. **the measurement, before the change** — what lengths exist, by read path. Two restrictions found in my own instrument (population; float tolerance) and recorded at the site.
2. **the instruments stop caring about the cell's shape** — the edit oracle and four sweeps routed through `isCellOn`/`cellOn` while the cell is still a boolean, so the product commit's numbers are attributable to the product.
3. **the cell carries the length** — model, three readers, four ops, the panel, the gate.

Part of #1010. Next: P4c — the printer preserves instead of re-deriving, where the 30 losses are predicted to go to zero.


MrityunjayBhardwaj:
## Critical self-review

Three findings, all in this diff. Two fixed here, one filed because it is P4c's decision rather than a bug.

**1. `resize` did not clamp lengths, while `quantize` and the whole roll do.** Fixed in `7f8c91a7`. Truncating a 4-column grid to 2 left a 3-column note reaching a column that no longer exists; downsampling pulled two hits together without shortening the first. Inert today (nothing reads a length until P4c), but a printer that then spelled it would produce notation the grid could not have meant. `clampLane` and `scaleCell` moved to `model.ts` beside the cell type in the same commit — the rule spans resize, resolution and (next) the printer, so it should not live in whichever file needed it first.

**2. P4a's comment said `atoms`/`spans`/`durs` would RETIRE in P4b. They did not.** Corrected at the site in `5d52db72` rather than left to rot, because a comment carrying a claim has no gate. The reason is worth keeping: what a *cell* needs is exactly that display view — one length and one anchor per distinct sound, first occurrence winning — so retiring them would re-derive the same dedupe at every call site. `occ` is still the column's truth; those three are its display projection. #1032 is where a column's two same-named notes stop being a choice at all.

**3. Filed #1045 — the writer's change-detection compares SOUNDS only.** `SourceRegion.content` for the grid is `GridCells` (`string[][]`), and `serializeStepGrid` compares it via `sameCell`. A model whose only difference from parse time is a note's LENGTH compares equal, so the region is copied back verbatim and the edit does nothing. Unreachable today — no gesture changes a length alone, and the ops that rescale lengths drop the regions via `restructured` — but P4c makes the printer honour lengths, and honouring them is undetectable while the comparison cannot see them. Noted at `sameCell` too.

### Also worth flagging on the diff itself

- **Two unreachable `?? 1` fallbacks**, in `columnsFromOnsets` (a `null` length, which only synthetic probe/anchor onsets carry) and in the leaf path (an anchor whose atom the engine did not play at that column, which `leaf-anchor-sweep`'s totality assertion says cannot happen). Both are commented as unreachable and both are covered by `cell-duration.test.ts`, so a wrong `1` fails rather than looking plausible. Left as fallbacks rather than throws because neither is a state the caller could act on.
- **`isCellOn` tests the shape, not `!== false`.** The exclusion form matches the type exactly and still admits a legacy `true`, whose `.duration` is `undefined` — that produced one `duration: NaN`. Nothing persists a `StepGridModel` (it is re-read from the document every time), so this guards the JS boundary and not a migration; stated at the site so nobody writes the migration that is not needed.
- **The A/B's first run lied, in my favour’s opposite direction.** It reported 10 differences that were entirely the instrument: the pre-change ops read `cell === true`, so handing them the new object cells made their upsample branch see an empty grid. A uniformly-empty arm is an instrument verdict, not a code verdict — the shape of a difference set is worth reading before its contents.

### Stack risk, unchanged and now one deeper

Seven PRs, none merged, and every body below `#1043` still quotes figures that `#1043` re-based. If they merge out of order or get squashed independently, those bodies become exactly the stale-figure problem this arc has been closing. This PR quotes only post-rebuild numbers.

Issue #1061, opened 2026-07-27
Grid coarsening is disabled in the UI, and two browser tests still assert the old contract


MrityunjayBhardwaj:
Found while landing #1048, on `studio_v0.2.0` @ `99fceeb6` — **not** introduced by that
branch. Attributed with a base arm: the same two tests fail on the merged tree with the
branch stashed.

## What's red

`packages/app/tests/resolution.spec.ts`, two of seven:

- `step grid: a lossless reduce keeps timing and is shown as a normal preset`
- `step grid: a non-power-of-2 (5-step) pattern can still be reduced (quantize)`

Both fail the same way — the preset they click is disabled:

```
locator resolved to <button disabled aria-label="4 slots"
  data-resolution-step="4" title="4 slots — unavailable">4</button>
```

The five roll/other cases in the same file pass. So this is grid **coarsening**
specifically.

## Why, measured through the real ops

Asking the shipped `stepSlotState` / `quantizeStepGridTo` / `scaleStepGridTo` directly:

| pattern | steps | target 4 | target 8 |
|---|---|---|---|
| `bd ~ ~ ~ sn ~ ~ ~` | 8 | **disabled**, both ops return the input unchanged | active |
| `bd ~ sn ~ bd` | 5 | **disabled**, both ops return the input unchanged | quantize |

This is the admissibility rule working exactly as intended, not a bug in it. Since P4c the
printer preserves a note's length, and a note lasting one column of an 8-column grid lasts
*half* a column of a 4-column one — which the grid has no notation for. The writer declines,
the op therefore returns its input, and the control is correctly disabled.

What is wrong is that **two shipped browser tests still encode the pre-P4c contract**, and
nothing re-ran them: they assert the button is enabled and that clicking it rewrites
`bd ~ ~ ~ sn ~ ~ ~` to `bd ~ sn ~`. P4c's own gates were all green, because the corpus gates
measure the ops and the printer, and this is the layer above both.

## The product question this actually raises

Reducing the slot count on a plain drum pattern is currently impossible in the UI. That is
the honest consequence of "a promise about lengths is a promise about room" — but a user who
types `bd ~ ~ ~ sn ~ ~ ~` and asks for 4 slots is asking for `bd ~ sn ~`, and today they get
a greyed-out button with no explanation.

This is the same boundary #1052 is about, approached from the opposite side. #1052 established
that **refining is a view change and coarsening edits your document**; this is what coarsening
[...]

Issue #1052, opened 2026-07-27
Should the step grid's resolution be a view/snap setting rather than a document rewrite?


MrityunjayBhardwaj:
Successor to #1051, which closed as not-planned once its premise was measured and did not hold. This is the question that turned up underneath it, and it is the larger of the two.

## The observation

Our `÷2` / `×2` / `slots` controls **rewrite the document**. Changing the step grid's resolution re-spells the pattern permanently.

In the products this UI is modelled on, resolution does not work that way:

- **Ableton Live** — grid resolution is a **snap** setting. Changing the grid changes the snap; holding Alt/Cmd bypasses it entirely; note data is never rewritten. Changing note length is its own explicit gesture (Shift + arrows), never a side effect of changing resolution. ([Editing MIDI](https://www.ableton.com/en/live-manual/12/editing-midi/))
- **Push 2's step sequencer** — the closest hardware analogue to our grid — adjusts resolution to change what the pads address, and exposes note length separately as per-note parameters. ([Using Push 2](https://www.ableton.com/en/manual/using-push-2/), §35.14.4)

So a user coming from a DAW reasonably expects the resolution control to change *how the pattern is displayed and snapped*, and expects their notes to be exactly where they left them afterwards.

## Why this is worth asking now

Three open problems all turn out to be the same problem seen from different sides:

1. **#1051** — coarsening declines where a note would go sub-column. It declines because it is being asked to permanently re-spell a pattern at a resolution that pattern does not fit. As a view setting there would be nothing to spell and nothing to decline.
2. **#1049** — `slots` and `×2` write *different note lengths* at the same ratio. Both rules are individually defensible; they only collide because each has to decide what to *write*. A view setting writes nothing, so the question does not arise.
3. The op-admissibility work in P4c (#1047) exists to keep controls from being enabled-but-inert. A large share of that surface is resolution ops. A non-writing control cannot be inert in that way.

## The honest counter-argument

Strudel's mini-notation has no separate "absolute time" model that a view could float above. A sequence's elements each take an equal share of the cycle, so the element count **is** the resolution — `"bd ~ sn ~"` is a four-element pattern, and there is no document-independent place to store "display this at 8".

So this is not a free change. It needs a real answer to: **where does the view's resolution live, if not in the text?** Plausible directions, none investigated:

- editor-local view state, persisted alongside the tab rather than in the document (nothing to write, but it does not survive sharing the file)
- a display hint in the source that the engine ignores
[...]

PR #1063, opened 2026-07-27
♻️ One definition of what a cell toggle does (#1048)

MrityunjayBhardwaj:
Closes #1048. The precondition for the resolution sprint's Phase 5 (#1058), which needs
somewhere for grid placement to live.

## The problem

The step grid's cell toggle existed in three copies:

| where | how it was written |
|---|---|
| `SequencerGrid.tsx:65` | `toggleCell` — the real one, what a click does |
| `edit-locality.test.ts:162` | `toggleCell` — a hand-written twin |
| `alternation-element-stress.test.ts:128` | inlined into the sweep loop, unnamed |

Worth noting for anyone auditing the rest: **only two are findable by name.** The third is a
bare `cells0[col] = isCellOn(...) ? false : cellOn()` followed by a `clampLane`, so a grep for
`toggleCell` returns two hits and reads as confirmation that there are two. It was found by
searching for what the code *does* rather than what it is called.

When P4c taught the toggle to clamp, that clamp had to be written three times — and nothing
would have failed had it landed in only two. The corpus copies would simply have simulated a
model no gesture can produce, and the writer would then have declined edits the UI clamps long
before they ever reach it. Which is the symptom that did show up, and it cost time to attribute
because the panel was already correct.

Two test copies of a production edit op are a second and third oracle for what an edit *is*.
They agree with each other and drift from the thing they measure, and a test that models the
edit cannot catch a change in the edit — it quietly keeps testing the old one.

## The fix

One definition, in `place.ts`, with the panel and both sweeps as callers.

**Why `place.ts` rather than `model.ts`,** which the issue suggested: `model.ts` owns the types
and the cell primitives (`cellOn`, `clampLane`, `scaleCell`) and holds no model-level ops at all
— those live in `lane.ts`, `place.ts`, `resize.ts`, `resolution.ts`. `place.ts` already owns
"put a note into a view and take one out, resolving what that costs the notes already there",
and the convention its neighbours keep is one module per concern with **both surfaces inside it**
(`resize.ts` = `resizeGrid` + `resizeRoll`; `resolution.ts` likewise). `place.ts` was the one
exception — roll-only — and that asymmetry is exactly the gap #1058 needs filled. So this adds
the missing half rather than a fourth module.

## Verification

**The acceptance criterion, proven rather than assumed.** Deleting the clamp from the single
definition turns both sweeps red, one failure each:

```
FAIL  alternation-element-stress.test.ts > grid: toggle every cell — no data loss, ≤1 element touched
FAIL  edit-locality.test.ts > grid: painting into one `,`-part does not rewrite the part beside it
```

Both therefore genuinely exercise the production op rather than a copy of it. Restored, green
again.

| gate | result |
|---|---|
| parity-corpus | 356 / 26 files ✓ (baseline) |
| editor | 3064 / 212 ✓ (baseline) |
| app | 962 / 71 ✓ (baseline) |
| tsc | 62 editor / 1 app — baseline, none in the touched files |
| sequencer browser suite | 13 / 13 ✓, including the real cell-click round-trip |
| editor `dist` | rebuilt and committed whole |

## Browser suite — attributed, and it turned up two things

The grid-touching specs showed 3 failures. Running the identical list on `99fceeb6` with this
branch stashed produced **the same failures**, so this branch introduces none. Pushing that
further, a serial re-run (which the config itself demands before calling any browser failure a
defect) gives **8 failures on the merged tree**, identical set on both arms.

Filed rather than folded in:

- **#1061** — the two `resolution.spec.ts` failures. Grid coarsening is disabled in the UI:
  since P4c preserves note lengths, a note lasting one column of an 8-column grid lasts half a
  column of a 4-column one, which has no notation, so the writer declines and the control is
  correctly disabled. The tests still assert the pre-P4c contract. **This is #1052's own subject
  from the opposite side** and the decision belongs there. Two measurements in the issue worth
  carrying into the sprint: the canonical defect reproduces exactly (`bd ~ sn ~` at 16 slots →
  `bd ~ ~ ~ ~ ~ ~ ~ sn ~ ~ ~ ~ ~ ~ ~`), and the two refine paths disagree about a note's length
  — `scaleStepGridTo` preserves it (`bd _ ~ ~ sn _ ~ ~`) while `quantizeStepGridTo` resets it
  (`bd ~ ~ ~ sn ~ ~ ~`), which sharpens #1049 with a mechanism.
- **#1062** — the other six, in full-song and timeline surfaces, all pre-existing. They were
  invisible because a full parallel run is genuinely noisy *and* the browser figure being quoted
  is a subset that excludes them.

## Test plan

- `pnpm --filter @stave/app exec vitest run tests/parity-corpus`
- `pnpm --filter @stave/editor test` · `pnpm --filter @stave/app test`
- `pnpm --filter @stave/app exec playwright test tests/sequencer.spec.ts`
- To re-check the gate fires: drop `clampLane` from `toggleCell` in `place.ts` and re-run the
  two sweeps.


MrityunjayBhardwaj:
Self-review pass — two findings, both handled.

**1. `toggleCell` landed in the notation layer with no unit test**, while its siblings in the same module (`placeNote`, `resizeNote`) have one. Added `place.test.ts`: six cases covering place, clear, lane isolation, the clamp firing, the clamp correctly *not* firing, and that a toggle cannot grow a lane past its own length. Proven to fire — removing the clamp reddens it, same as the two sweeps. Editor suite 3070/213.

Writing it turned up something worth recording: **a hand-built model is not a realistic one.** The first draft replaced `lanes` wholesale on a parsed model, which broke its source anchoring — the clamp assertions passed while serialize returned `null`, a decline. The cases now start from a real source (`bd _ ~ ~`), so they assert what a click on that pattern actually produces.

**2. Routing the alternation sweep through the production op is not a pure substitution.** The old inline code wrote `cells0[col] = cellOn()` by direct index, which *extends* the array when `col >= cells.length`; `toggleCell` maps over existing cells and cannot. That could have silently shrunk the sweep's coverage, so I measured it rather than reasoning about it: **0 of 966 parsing models have a lane shorter than its grid** (57 in the alternation sweep's own population, also 0). The two are therefore equivalent on this corpus, no coverage lost, and `cells.length === steps` is a real invariant — now pinned by one of the new cases. The counts are non-zero on both arms, so this is a measured zero rather than an empty loop.

Worth noting the direction: the panel's `paintCell` already guards past-the-end, so the old sweep was writing a model no gesture can produce. That is the same defect class this PR exists to close, one instance smaller.

PR #1067, opened 2026-07-27
🔬 Spike: placing a hit subdivides only the element under it (#1058)

MrityunjayBhardwaj:
The locality spike for **#1058**, run before Phase 1 (#1054) because it could invalidate the rest of the resolution sprint (#1052). Additive only — **no product code changes**, two new files under `tests/parity-corpus/`.

## Problem

#1058's case for the sprint rests on one property: placing a hit at a position the grid cannot yet express must subdivide **the element under the cursor and nothing else**. If placement frequently needs more than that element, the benefit collapses from *"refining is free AND edits stay local"* to just *"refining does not write"* — still real, much smaller, and #1054–#1057 need re-scoping.

Asserting the property from the model would prove nothing. It had to be asked of the shipped reader, the shipped writer and the real engine, with each property gated separately — and the result had to be un-readable-to-taste.

## Fix

**The decision rule is committed first, in its own commit (`6901bb8d`), before the probe existed.** Population, gesture, the three properties, the GO / RE-SCOPE / INVALIDATE / HALT thresholds, and the one pre-existing change the play-check is allowed to see (the placement clamp shortening a note that was sounding through the new onset, so a correct clamp cannot be mistaken for corruption).

The probe uses shipped code throughout — `parseStepGrid`, `place.ts`'s `toggleCell`, `serializeStepGrid`, and the committed `engineEditOracle`. The only new code is a pure rescale of the model and the source regions that index it, which is the machinery the phase would ship. It is a `.spec.ts`, so vitest's `include` never collects it: a probe, not a gate.

## What it found

819 corpus units on the element-splice path, 15,212 placements.

- **The precondition holds** — `bd [~ hh] sn ~` opens at 8 columns and round-trips byte-identical today.
- **It is local: zero non-local writes.** And it needs no new writer — the region splice already spells a multi-column step as a group, so the subdivision falls out of machinery shipped since #913.
- **It plays** — 86 corrupt asks over 3 units, and all 86 also corrupt at the shipped resolution with no subdivision. This mechanism introduces none.
- **Acceptance is 40.9%, and it measures something else.** 98.6% of declines are one mechanism; a base arm with no subdivision reads 85.0% with that same mechanism at 98.2% of its declines, and 99.0% when the clamp spans the column instead of the lane.
- **Nesting depth stays at 1 and never deepens.** The group widens instead, and the sequence terminates when the reader refuses a document the writer just emitted.

**Verdict, both readings recorded rather than reconciled:** by the letter of the rule, INVALIDATE (acceptance < 50%). By the evidence, the trigger is a pre-existing and separately-fixable gap, while the property the rule was aimed at passes at every depth. Recommendation is GO gated on #1064; the call is the reviewer's.

Filed from the run: **#1064** (the per-lane clamp, with the product question it carries), **#1065** (a chord's duplicate member lost on edit), **#1066** (the writer emits notation its own reader refuses).

## Test plan

- `pnpm --filter @stave/app exec vitest run tests/parity-corpus` — **356 passed / 26 files**, the committed baseline, unmoved. The probe is a `.spec.ts` and is correctly not collected.
- The probe itself runs explicitly against a local config; its output is transcribed in `1058-LOCALITY-SPIKE.md` under FINDINGS.
- Attribution: every corrupt ask and the whole decline population are measured against a base arm on the same units with no subdivision, so nothing here is credited or blamed on the mechanism without that control.

Refs #1058 · #1052 · #925. Does **not** close #1058 — the phase itself is unimplemented and now gated on #1064.

### DIFF
--- a/packages/app/tests/parity-corpus/1058-LOCALITY-SPIKE.md
+++ b/packages/app/tests/parity-corpus/1058-LOCALITY-SPIKE.md
@@ -261,6 +261,32 @@ with a control set — the cause is the accumulated `_` sustains, not the width:
 A writer emitting notation its own reader will not re-open is a round-trip hole. Filed
 separately; the mechanism is **not** diagnosed here and this doc does not guess at it.
 
+### What this spike did NOT measure — stated so no one reads it as covered
+
+- **Only `k = 2`.** #1059's presets are absolute counts (4/8/16/32/64), which are `perBar × k`
+  for larger `k`. A larger `k` produces longer runs of `_`, which is exactly what #1066's
+  refusal is sensitive to, so acceptance and re-openability at `k > 2` are **unknown**.
+- **`MAX_STEPS = 64` is bypassed here.** The refine rescales the model directly rather than
+  re-reading, so a unit at 40 columns refines to 80 and is measured — while the reader would
+  refuse to re-open it. That interaction is #1055's question and is **not** answered here.
+- **The two population restrictions cost nothing, which is checked rather than assumed.**
+  819 + 569 + 82 + 57 = 1527 exactly: zero units were dropped for a non-identity base, zero
+  for a span-walk mismatch, zero for a null refine. The span walk reproduced the original
+  mini byte-for-byte on all 819, which is what makes the element offsets sound.
+- **The hypothesis arm is judged against a rule written for the shipped op.** Its 1,293
+  "play-differs" asks are the cross-lane clamp firing as designed, not defects — the
+  pre-registered allowance only covers a same-lane clamp. So "90.5% plays" reads that arm
+  conservatively.
+
+### One note on the instrument, because it nearly produced the opposite verdict
+
+The probe's first version compared **refined** columns against **unrefined** region ranges,
+so every click was attributed to the next element along. It reported `L = 21.2%` — a clean
+INVALIDATE — and nothing in it failed. It was caught by printing the non-local samples and
+reading them: `C D`, click on element 0, reported as element `D`. The fix moved `L` from
+21.2% to 95.7% with no change to any product code. A locality measure that indexes its own
+reference frame wrongly fails in the direction that looks like a finding.
+
 ---
 
 ## VERDICT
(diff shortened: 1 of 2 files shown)


MrityunjayBhardwaj:
## Self-review pass

**One real gap found and fixed (`bd50…`, pushed).** The probe read `.model` off `parseStepGrid`'s return value on both arms of its discriminated union — the refusal arm has no `model`, and it is only invisible because JavaScript hands back `undefined`. It moved the app package's type-check from **1 error to 9**. The vitest suite was green throughout and said nothing about it, because vitest does not run tsc.

Fixed by narrowing on `ok` through one helper at every site. The refusal *reason* now comes from that helper instead of a cast, which is what lets the control-set rows in the findings print it at all.

**Verdict-neutral, observed rather than assumed** — every figure re-runs identical: 0 non-local writes, acceptance 40.9% / base arm 85.0%, corrupt 86 over 3 units, all four sampling depths unmoved.

### Gates, re-run on the final tree

- `tests/parity-corpus` — **356 passed / 26 files**, the committed baseline
- `tsc` app — back to **1** pre-existing error (`edit-locality.test.ts:120`), from 9

### Two limits recorded in the findings rather than left implicit

- **Only `k = 2` was measured.** #1059's presets are absolute counts, i.e. larger `k`, which produces longer runs of `_` — exactly what #1066's refusal is sensitive to. Acceptance and re-openability above ×2 are unknown.
- **`MAX_STEPS = 64` is bypassed here**, because the refine rescales the model instead of re-reading. A 40-column unit refines to 80 and gets measured, while the reader would refuse to re-open it. That interaction is #1055's question and is not answered by this run.

### Also on the record

The probe's first version mis-indexed its own reference frame — refined columns compared against unrefined region ranges — and reported a clean INVALIDATE at 21.2% locality with nothing failing. It was caught by printing the non-local samples and reading them (`C D`, a click on element 0 attributed to element `D`). No product code was involved either way. A locality measure that gets its frame wrong fails in the direction that looks like a finding, which is the direction nobody double-checks.

PR #1071, opened 2026-07-28
🍃 A cell is offered only where the writer will take the note (#1064, #1070)

MrityunjayBhardwaj:
Carries the two decisions taken on #1058's spike: **#1064 → refuse now, clamp later**, and **#1070 → suppress the placement affordance wholesale on leaf views, keep delete/edit**.

Three commits: the fix, plus two things self-review found in it.

## The problem

#1010 P4c established that an enabled control and a working one are the same claim, and swept it corpus-wide — 20,244 op applications, 0 unwritable. It reached every op and every control-state function. It never reached the **cell**, which is the gesture the panel exists for.

`SequencerGrid` toggled unconditionally; `useGridModel` returns on a null serialize *before* updating the model. So a click the writer could not spell wrote nothing, toggled nothing, and said nothing.

Measured over the mini-corpus with the shipped readers, ops and writers, split by write path:

| surface | path | units | placements | inert | |
|---|---|---:|---:|---:|---:|
| grid | element | 819 | 11,633 | 1,748 | 15.0% |
| grid | **leaf** | **82** | **3,584** | **3,584** | **100.0%** |
| grid | alt | 57 | 3,427 | 512 | 14.9% |
| roll | **leaf** | **54** | **18,386** | **18,386** | **100.0%** |

The leaf path is a different cause and the dominant one — 88% of all inert placements. It writes by byte surgery at each note's own span, so it edits and deletes what exists and can never create: **a new note has no span**. Deletes on the same units go through (63 written vs 0 placements), so this is unsupported *by construction*, not broken.

## The fix

**One rule, in the op.** `toggleCell` and `placeNote` ask the real writer before returning — the same `ifGridSpellable`/`ifRollSpellable` gate `resize.ts` and `resolution.ts` have applied since P4c. A refusal is the input by reference, which is what the `notation/` family already means by "could not apply": `mutate` skips the write, and `canToggleCell`/`canPlaceNote` are *derived* rather than a second predicate that can drift.

It lives in the op rather than the panels deliberately — enumerating an admissibility rule caller-by-caller is exactly how the cell was missed in the first place.

**`viewPlacesNotes` answers the path question once per view.** On a creation-incapable path the honest answer is "this view does not place notes", not 3,584 individual declines, so the panel states it once and leaves the cells legible.

Untouched: deletes, resizes, velocity drags, ⌘-click paste-target selection.

## What self-review found in it

**A refused paste deleted its own target** (commit 2). The roll's paste is a clear composed with a place. Giving `placeNote` a refusal path that returns its input meant a refused paste returned the model with the target note *already removed* — and that serializes fine, so the write went through. Measured: **573 (note, duration) cases** across the corpus, on the alt path as well as leaf. Pasting a 3-step note over the `4` in `<4 6 3 [6 7]>` would have written `<~ 6 3 [6 7]>`. Fixed by making replace-at-target one op (`pasteNote`) that refuses by returning the *original* model. Composing a decline-capable op with one that cannot decline puts the whole gesture's atomicity on the caller, and the caller got it wrong.

**The roll's per-cell map was not cheap** (commit 3), which the first commit asserted in a comment without measuring. Per model change, and `mutate` fires on every frame of a move drag:

| surface | p50 | p90 | p99 | worst |
|---|---|---|---|---|
| grid | 0.01ms | 0.19ms | 2.08ms | 14ms / 608 cells |
| roll | 0.28ms | 2.63ms | 21.67ms | 50ms / 1,632 cells |

So the roll asks the view-level question only — 96.3% of its inertness is the leaf path, caught by one boolean at no cost — and the grid keeps its map, where p99 is 2.1ms and it catches 15%. Correctness is identical either way; only the affordance for the roll's residual 0.9% differs, and that is unchanged from today. Filed as #1072 rather than left implicit.

## What did not change

- **Writer reach.** #986 raised it onto the leaf path (grid 80→131, roll 56→85). Reach asks "can this write back?", which every leaf view answers yes. Acceptance of *new* content is a different axis; those figures are true and unaffected.
- **The refusal set.** The alternation sweep's pinned 512 declines reproduce *exactly* at the op, with the writer now declining zero — same refusals, observed early enough to offer.
- **Every non-leaf view.** All 468 non-leaf grids with an empty cell still offer at least one placement — 468 of 468, pinned.

## Test plan

New `placement-admissibility.test.ts` (parity-corpus). It states its population, comparison axis and sampling depth, reproduces the committed per-path figures exactly, and compares the offer against an **independent ungated arm** — `canToggleCell` is derived from `toggleCell`, so asking whether they agree would be a tautology that reads green forever.

Every gate proven able to fire by breaking it: removing the grid gate → disagreements + leaf views offering; removing the roll gate → roll disagreements; forcing `viewPlacesNotes` true → path assertion red; removing the paste guard → both new assertions redden and the failure prints the lost note.

Observed through the real gesture on a real document, not forced state: a leaf grid renders the statement, its empty cells are genuinely disabled (Playwright itself refuses the click), a forced click still leaves the document byte-identical, and a held note still deletes. An ordinary grid shows no message, no inert cells, and places normally.

| gate | result |
|---|---|
| parity-corpus | **359**/27 (was 356/26) |
| editor | **3079**/213 (was 3070) |
| app | **965**/72 (was 962/71) |
| tsc | **62** editor / **1** app — baselines |
| sequencer + piano-roll + projection, serial | **43/43** |

⚠ Browser figures are the three specs this touches, run with `--workers=1`. The suite-wide "53/53" is a subset and is not quoted — #1062.

One population note stated rather than pinned away: the new gate counts 827 parseable element-path units where the committed sweep counted 819 placeable ones. The 8 contribute zero asks (a grid with no empty cell has no placement to offer), which is why every ask-level figure is identical.

Refs #1064 · #1070 · #1072 · #1058 · #1052 · part of #925.


MrityunjayBhardwaj:
One thing found while building the follow-up, worth knowing before this merges — it does not change the diff, and I do not think it should block.

Giving `toggleCell` a refusal path that returns its input silently disabled a measurement in a committed artifact. `_1058-locality.spec.ts` decides acceptance with `serializeStepGrid(toggleCell(...)) !== null`, which was sound while an unspellable placement produced an unspellable model. Now a refused placement returns the user's current model, and that serializes fine — so the metric counts refusals as acceptances and reads 100.0% at every depth against the committed 40.9%.

It is the same shape as the paste bug this PR already fixes: the refusal signal means something only to a caller that passed its own current model, and a consumer that reasons about the *result* instead of the *identity* silently changes meaning. The paste was a caller; this is an instrument. Both were invisible because the change reads as purely restrictive.

Filed as #1073 with the observed figures and the one-line fix (`next === input`). No product code involved, and nothing here needs to change.

PR #1075, opened 2026-07-28
📏 A note is drawn across the columns it covers (#1056)

MrityunjayBhardwaj:
Resolution Phase 3 of #1052. **Stacked on #1071** — base is `fix/1070-1064-placement-admissibility`, not `studio_v0.2.0`, so review that one first. `Closes` never fires on a non-default base; #1056 gets closed by hand.

## The gap

The step grid could not show how long a note lasts. Every visual property of a cell derived from `isCellOn`, and a column a note is *carried through* is `false` — so `bd _ sd ~` (bd sounds for two columns) and `bd ~ sd ~` (one) drew the same picture.

What makes this worth stating carefully is how much of the stack read TRUE while the user saw nothing. The length was read by the parser (#1010 P4b), preserved by the printer (P4c), and asserted per cell against the engine (`cell-duration.test.ts`). None of that is displayability — that is a property of the rendered geometry and of nothing upstream of it. Which is why it mattered: it made note length an axis an op could change as collateral with nobody able to perceive the change.

## The change

`laneCoverage` (`notation/model.ts`) derives, per lane, which columns each note occupies and by how much of each. It is a **read**, not an op, which is why it sits beside `clampLane` rather than in `place.ts` — and it asks `clampLane`'s own room rule non-destructively, so the drawing and the document cannot disagree about what is spellable.

The panel then draws a bar whose **width** is that extent and whose **height** stays velocity — two orthogonal axes on one bar, the way a DAW draws a note. A carried column is dimmed rather than solid, which is the vocabulary the piano roll already ships (`opacity: on && !isHead ? 0.7 : 1`); one held note reads as one note, never as a second trigger.

**A note lasting exactly its own column is the solid square it has always been.**

Also removes the defer-comment at `SequencerGrid.tsx:343`, which named #1010 P4d as the owner of this work. P4d is grid leaf-first writer routing and owns nothing about rendering.

## Observed, not only verified

The decisive pair is the one recorded as identical when the gap was found. In the live app:

```
bd ~ sd ~   →  #---
bd _ sd ~   →  #=--      head opacity 1 · carried opacity 0.7
```

and `<bd - - -> *2` draws a genuine **half-width** bar — the sub-column case, which is the one thing the roll never needs.

Control arm in the same browser run: an all-length-1 pattern has 0 carried columns, 0 narrowed columns, and a fill width of exactly `100%`.

## Two findings from measuring instead of reasoning

**A float sliver is not a column.** Lengths reach the cell through Fraction→float, so "exactly one column" routinely arrives as `1.0000000000000004`, and the first draft claimed the next column at an extent of `4e-16` — an invisible bar in a column the pattern never sounds through. 61 phantom columns corpus-wide. They decompose with no residue (carried 1550→1489 and partial 73→12 are the same 61), and **all of them are on the derived paths while the syntactic path loses exactly zero** — which is the attribution that says float noise rather than length.

**The roll is only half the no-op it was predicted to be.** #1056 asked for that to be measured rather than inferred, and it was right to. The coverage walk needs nothing — 4825 of 4842 notes — but 17 across 4 minis carry a fractional start or length its integer walk cannot draw, and **7 of those render in no column at all**. Pre-existing and untouched here. Filed as #1074, with the split pinned in this PR's gate so it cannot drift silently.

## Gates

New `cell-coverage.test.ts`, reporting the two surfaces separately, with the circularity stated: on the **syntactic** path (3931 columns) lengths come from AST spans and never touch the engine, so the engine comparison is real evidence; on the derived paths it would be restatement, and is labelled as such.

Every arm proven to fire by breaking the rule three ways — reverting to one-box-per-trigger (3 red), removing the sliver threshold (2 red), decoupling the head from the trigger (5 red).

Structural control: **heads 4718 = ON cells 4718**, so no trigger moves, appears or vanishes. carried **1489** · partial **12**, both pinned.

| gate | result | baseline |
|---|---|---|
| parity-corpus | **366** / 28 | 359 / 27 (+7 tests, +1 file — this gate) |
| editor | **3079** / 213 | 3079 / 213 |
| app | **972** / 73 | 965 / 72 |
| tsc | **62** editor / **1** app | 62 / 1 |
| sequencer-note-length (serial) | **3/3** | new |

## Not done, and blocked rather than skipped

The issue also asks for the k = 2 and k = 4 arms. The view multiple is #1055's parameter and #1055 is not shipped, so there is no k to run at — everything here is k = 1. The comparison is resolution-generic; re-run the file with the multiple once #1055 lands.

## What this unblocks

The #1064 phase-2 column clamp, parked on `spike/1064-phase2-column-clamp`, was held for one reason: it shortens a note in a sibling lane on 2,229 placements and the grid could not draw the axis it was changing. It can now.

Part of #1052 · #925.


MrityunjayBhardwaj:
## Critical self-review

Two gaps in my own diff, both fixed in `95bd16f4` rather than filed.

**A richer picture left one audience behind.** A carried column became visually distinct and its accessible name stayed identical to an empty cell's — so the change was legible to sighted users and absent for everyone else, which is the same failure this phase exists to fix, one audience over. It reads as purely additive (a fill appears, nothing is removed, everything stays green), and the regression is in what *didn't* change. Now names where it is held from, asserted in the browser spec beside the fill it explains. No test asserted the old label — checked with a control arm, since a zero-hit grep is not absence.

**The new memo's comment claimed it was cheap.** That is the exact shape of the claim that was wrong the last time a per-cell map here was called cheap: an unmeasured comment, believed because it sits next to the code. Measured instead — p50 0.0006ms, p99 0.0063ms, worst 0.0985ms over the 958 corpus models, against `placeable`'s p99 2.25ms / worst 13.10ms on the same run, 0.86% of its total. The claim held. It had no business being made without the number.

**And one thing I reasoned wrongly and the measurement corrected:** I had concluded a carried column would be *inert*, since placing into an occupied column is what #1064 counts as refused. It isn't — `canToggleCell` returns true, because the per-lane clamp shortens the previous note and the write succeeds; #1064's refusals are cross-lane. Had I acted on the reasoning I'd have shipped a tooltip explaining a refusal that doesn't happen.

## What I deliberately did not do

- **Fix the roll.** Its 17 fractional notes are a real defect and pre-existing; widening this phase to cover them would mix a rendering change with a parse-adjacent one on a surface I'd have had to re-baseline. Filed as #1074, with the split pinned here so it can't drift silently. The fix wants `laneCoverage` shared across both surfaces, which is #1032's direction and not a rename — the roll's *start* is fractional too, which the grid's column-indexed model has no equivalent of.
- **The k=2 / k=4 arms.** They need #1055's view multiple, which isn't shipped. Stated in the gate's header as blocked, not skipped.

## What this gate cannot catch

The syntactic arm is real evidence; the two derived arms are **partly circular by construction** — coverage descends from the same haps the expectation does — and are labelled that way rather than pooled into one green. What they still check is the column placement and the cycles→columns conversion.

The structural control (`heads 4718 = ON cells 4718`) proves no trigger moved, appeared or vanished. It says nothing about whether the *extent* is right; that is the syntactic arm's job, and it covers 3931 of 6207 columns.

MrityunjayBhardwaj:
## Both caveats are closed — `c13d0798`

**The k=2 / k=4 arms now run.** They were reported blocked on #1055's view multiple; that was one option too many. `scaleStepGrid(m, 'double')` already ships and already doubles both the column count and every cell length — exactly the input a finer view would hand the renderer. The property asserted is that what the user SEES does not change: each note's drawn time doubles in columns and every head lands on twice its index. **876 units at 2×, 876 at 4×**; the 82 the writer cannot spell at 2× are counted, not filtered away in silence.

**The circularity is broken by comparing against the writer, through text.** `sustainTokens` answers the same covered-columns question in its own loop and spells the answer `_`, so the serialized document is a source `laneCoverage` had no hand in — on the derived paths too.

Getting there took three tries, all corrected by measurement rather than review:

1. **The first version was a tautology.** It round-tripped the model and compared the picture before against after — building *both* with `laneCoverage`, so breaking the coverage rule broke both sides identically. It read green while four other arms went red. Caught only by running the red-test and asking which arms did **not** fire. That is now a catalogued discipline: a red-test is evidence only for the assertions that failed by name.
2. **The replacement tested flatness by counting tokens**, which admitted `[a4,c#4,e4],[~ f#6 e6]` — it yields `steps` pieces and the pieces are fragments like `e6]`. Flatness is a property of the syntax, not of a count that correlates with it.
3. **Restricted to single-lane models it compared 357 units containing zero `_`.** `bd _ sd ~` is *two* lanes sharing one token sequence — sustain is a per-**column** fact, and the per-lane restriction had defined the subject out of the population.

**What it still cannot see, pinned rather than glossed:** the flat population holds **0 sustains at 461 units** — every carried column in this corpus lives in a grouped or stacked mini, which the arm excludes because a column is not one token there. So at corpus scale it proves only that the grid never draws a carry the document lacks. That count is asserted `=== 0` with a note to widen it if the corpus changes, and the *other* direction is covered by a fixture arm on minis that do carry a sustain. **Both directions proven to fire** — under-draw reddens 5 arms, over-draw reddens 7.

parity-corpus **369**/28 · tsc app **1**.

## The followup is also done — #1074, in #1076

Stacked on this branch. The roll's 17 fractional notes are fixed by giving both surfaces **one** rule (`columnOverlap`), and the roll arm here re-pins to **4842/4842 exact, 0 invisible, 0 misdrawn**. It also fixes a second oracle in this very gate: the roll arm had re-implemented the panel's integer predicate inline, and agreed with it only because both were wrong the same way.

PR #1079, opened 2026-07-28
🎯 The selection ring is drawn above the note, so selecting one shows something (#1077)

MrityunjayBhardwaj:
Selecting a note in the piano roll showed nothing. The ring was there in the DOM the whole time — it was painted underneath the note.

### The defect

The ring is an inset `box-shadow` on the cell button. CSS paints an inset shadow with the element's background, so every child paints above it. That cost nothing while the note **was** the background; #1076 moved the note into a `<span data-roll-fill>` so a fractional note could occupy part of its column, and the fill is opaque and covers the whole column for an ordinary note.

So `data-roll-selected="true"` was set, the computed `box-shadow` was correct, the gesture worked — and the user saw nothing. A selected **empty** cell was never affected, because there is no note to cover its ring.

### Measured, against a real base arm

Near-white pixels in one cell's 2px inner frame, 4× device scale, driving the real app. The base arm is the previous commit's committed `dist`, checked out and served — not a reconstruction.

| arm | pre-#1076 | shipped `27146b55` | this branch |
|---|---|---|---|
| selected cell **with a note** | 424 / 2300 | **0 / 2300** | 403 / 2300 |
| selected cell, **empty** (control) | 424 / 2300 | 424 / 2300 | 392 / 2300 |
| note cell, not selected (control) | 0 / 2300 | 0 / 2300 | 0 / 2300 |

Both controls are flat across the regression, so the note fill is the whole difference. The ring now sits a pixel further in (it is inset by the border rather than drawn in the border box), which is the small drop from 424.

### The fix

The ring moves off the button's style and becomes the cell's **last child** — an inset-shadow overlay. Paint order inside a cell is DOM order, so being last is what keeps it above the note, and it frames whatever is underneath — note, name, resize handle or bare cell — without having to know which.

### The playhead needed nothing, and that was checked rather than assumed

#1076 claimed both the playhead border and the selection ring stay at full strength. The playhead half holds, and for a real reason: it is a `border`, painted in the border box, while the fill is absolutely positioned against the padding box. With the transport running — 77.4% white frame pixels on a playing cell holding a note, 77.6% on a playing empty cell.

### The gate

`piano-roll-selection-ring.spec.ts` compares a cell against **itself**, selected and not, so it needs no reference image and no image library. It asks about pixels deliberately: every existing assertion about the ring asks the DOM, and `data-roll-selected` was true throughout the regression.

Three tests, one claim each — a failing assertion ends the test, so a second assertion after it is evidence for nothing. **Proven by breaking it**: with the fix reverted, the appearance test and the paint-order test both go red by name, and the empty-cell control stays green. That control is the arm showing the harness can see a ring at all; if it ever fails alongside the others, the harness is broken rather than the ring.

### Gates

editor **3079**/213 · parity-corpus **369**/28 · app **975**/73 · tsc **62**/**1** · piano-roll + note-length + copy-paste + sequencer browser **46/46** serial, including all 23 roll gesture tests.

Note geometry re-observed on the built app after the change and identical on every fixture — control melody, held note, chord, the fractional fixture, black keys, velocity colour mode. The overlay renders only when a cell is selected, so nothing else could move.

### Not in this PR

#1078 — the same root showing up elsewhere: on a note that does not fill its column, the note **name** and the **resize handle** are still positioned against the cell, so they end up off the bar. 12 and 18 notes respectively, over the same 4 minis. It carries a design question about how small a resize handle may get before it is unaimable, which is why it is not bundled here.

Found while testing the roll after #1076 — the 26/26 that shipped with it asserted round-trips and gestures, which is behaviour, and the change was about paint.

Closes #1077.


MrityunjayBhardwaj:
Self-review of the diff, after the PR was up:

- **Measured a proposed improvement and dropped it.** The overlay is laid out against the padding box, so the ring sits one pixel further in than the inset shadow it replaces (403 white frame pixels vs the original 424). Growing it by the border width to close that gap is the obvious move and is worse — at `inset: -1` the ring draws outside the cell and bleeds across the 1px lane gap, 2156 of 2300. Second commit records the number in the comment rather than changing the geometry.
- **Cost is zero per cell.** The overlay renders only for the one selected cell, so this adds no per-cell work to a surface where that has bitten before.
- **The fragile part is the part the gate asserts.** "Last child" is what keeps the ring above the note, and it stops being true the moment someone appends another child — which is why there is a test for the DOM position and not only for the appearance.
- **Checked the sibling surface rather than assuming.** The step grid draws a fill the same way, but its cell has no ring, name or handle to cover — the fill is its only child — so the class does not exist there today.
- **Gap found, filed as #1080:** selection has no accessible signal at all. `data-roll-selected` and a ring are a data attribute and some pixels; `aria-current`/`aria-selected` appear nowhere in the panels. Pre-existing, and worth deciding alongside whatever role the grid should declare rather than bolting an attribute on here.

Still open from the same root: #1078 (the name and resize handle are cell-anchored, so they sit off a fractional note's bar).

PR #1081, opened 2026-07-28
🎯 A new hit ends every note sounding through its column, across the part (#1064 phase 2)

MrityunjayBhardwaj:
Phase 2 of #1064, unparked. The clamp was built and measured last session and deliberately held back on one condition; #1056 discharged it, and this PR confirms that by observation rather than by inference.

### The defect

`toggleCell` clamped the **lane** it edits. The writer's constraint is per `,`-**part** — one token per column, a sustain spelled `_`, `[_,bd]` unspellable. So a note sustaining in a *sibling* lane blocked the column, the model handed to the writer said two things at once, and the click went inert with no feedback. That was **1,717 of the element path's 1,748 declines (98.2%)**.

`clampPartAtOnset` ends every note still sounding through the new onset's column, across the part and no further. The part boundary is where the writer's constraint stops — parts serialize independently, so `bd _ _ _, ~ ~ hh ~` is legal and needs no resolution — and trimming past it would shorten a note the writer was always going to accept.

### Why it ships now, observed rather than assumed

It was held because the step grid drew a two-column note identically to a one-column one: the collateral trim was invisible, and a view that cannot display an axis must not change it. #1056 made length drawable. That is checked here in the app, not inferred from "#1056 shipped":

`s("bd hh*2 sd cp")` opens with the kick covering two columns. Clicking the hi-hat at column 1 writes `[bd hh] hh*2 sd cp` and the kick's bar goes **108px over 2 columns → 54px over 1**. Corpus-wide, **2,229 of 15,029 accepted placements** trim a sibling lane, and every one of them is now drawn.

### Re-measured on the rebased tree, not carried over

The parked figures were measured before three merges, one of which rewrote the roll's rendering. Everything below is a fresh sweep. Asks are identical on all three paths — an ask is an empty cell, and the clamp neither creates nor destroys one — so the refusal column is the whole delta:

| path | refused before | after | asks |
|---|---|---|---|
| element | 1,748 (15.0%) | **31** (0.27%) | 11,633 |
| alt | 512 | **0** | 3,427 |
| leaf | 3,584 | 3,584 | 3,584 |

Leaf is unchanged and must be: byte surgery has no span to create, so #1070 is untouched by this.

### The residual is now asserted, not described

The 31 are the residual #1064 named in advance. They were classified in a comment; they are now a test: **31 refusals over 9 minis, every one on a multi-part unit**, with the **64** multi-part units in the population pinned alongside so the claim cannot quietly become a statement about an empty set. It is its own `it()` — an assertion after a failing one never runs, so folding it into the sweep would have made it evidence for nothing on exactly the breaks that matter.

### Two inherited comments were wrong, found by measuring them

- **"and NOT ONE is fractional" reported on an empty set.** The element path holds *zero* units with a fractional cell length, so no refusal could have been fractional. Clause dropped. (The detector was checked against #1069's own example, `[bd _ _ _, ~ ~ hh ~]`, which does read fractional — `bd@0.5 sd hh cp` is the wrong control, since the reader raises the resolution and hands back integral lengths.)
- **The 8-unit gap is duplicate corpus entries**, not "parseable models vs models that open a placeable view". `mini-corpus.json` holds 1535 entries and 1527 distinct minis; this sweep maps them raw, the committed one dedupes. The stated *consequence* was right (they contribute zero asks) while the *reason* was not, which is what let it survive.

### `alternation-element-stress.test.ts`: 512 → 0

That gate pinned 512 alt declines and the 18 minis producing them. Both go to zero, and that is the edit succeeding rather than a guard being weakened: the counterfactual behind those 512 established they differed in **duration only**, and the refusal was about the shortening being *invisible*. The 18 minis are kept in the comment so the change stays legible, and the set is still pinned exactly rather than by length.

⚠ Worth noting for review: this test was never updated on the parked branch, so the spike would have been red on the full parity suite. It only surfaced here because the rebase was followed by a full run rather than by the one gate the change touched.

### Proven by breaking it

With the clamp made a no-op, the per-path counts arm and the residual-classification arm both go **red by name**. Three arms stay green, and that is expected rather than overlooked: the roll arm does not use this clamp, and the "accepted placements leave nothing sustaining" arm inspects accepted placements only, so a disabled clamp turns them into refusals it skips — its own comment says so.

### Decisions already taken, not re-litigated

Clamp **uniformly**, not conditioned on audibility · **one rule, both surfaces** · the roll has always clamped across the column, so "both surfaces" meant the grid moving to meet the roll.

### Gates

parity-corpus **371**/28 · editor **3082**/213 · app **977**/73 · tsc **62**/**1** · sequencer + piano-roll + note-length + selection-ring + copy-paste browser **46/46** serial.

Closes the phase-2 half of #1064. #1058 was GO gated on this.


MrityunjayBhardwaj:
Self-review of the diff, after the PR was up:

- **Consolidated an epsilon the rebase exposed** (second commit). `clampPartAtOnset` decided "is this note still sounding through that column?" with its own `1e-6`; it was written before `columnOverlap` existed, so after the rebase the file answered one question at two thresholds a hundred lines apart. It now asks the shared rule. Measured first, because a threshold change is a behaviour change until shown otherwise — parity-corpus 371/28, editor 3082/213, and every pinned per-path figure identical. No behavioural component.
- **The rebase was the stacked-squash shape**, not a fast-forward: the spike sat on the pre-squash #1071 branch, so only `cdf2421e` was replayed (`rebase --onto`). Conflicts were `model.ts` (a clean union — both sides added functions at the same point) and `dist` (taken from base and rebuilt).
- **Deleting a cell still skips the part clamp**, deliberately: erasing an onset can only give the notes around it more room, never less.
- **`clampLane` is still called and is not redundant.** The part clamp resolves the conflict at the new onset; `clampLane` maintains the room invariant across the whole lane. Different scopes.
- **Nothing was carried over from the parked branch.** Every figure in the description is a fresh sweep on the rebased tree, and the acceptance condition was discharged by watching the kick's bar halve in the app rather than by reasoning from "#1056 shipped".

Open and unchanged by this PR: #1069 (fractional column lengths, deferred) owns the one class the residual comment used to claim was absent.

PR #1091, opened 2026-07-29
🎹 The roll draws every note it carries, and the velocity lane only offers a drag it can write (#1087, #1089)

MrityunjayBhardwaj:
Two piano-roll defects that share a shape: the panel showed or offered something the model could not back up, and said nothing when it fell through. Both were found by measurement rather than by report, and both ship with the population they touch recorded as a committed arm.

Stacked on #1090 — #1087 extends the column-rule home that #1085 created, and #1089 resolves a deferral #1086 wrote into a comment.

## #1087 — a fractional `@n` weight emptied the panel

`@n` is a relative **weight**, so a pattern's length need not be a whole number of columns: `note("c4@1.5 e4@1.2")` is 2.7 columns long. The panel counted its cells with `Array.from({ length: model.steps })`, and `Array.from` floors its length.

```js
$: note("c4@0.2 e4@0.2 g4@0.2 b4@0.2 c5@0.2").sound("piano")
```

The weights sum to `0.9999999999999998`. The Pattern tab bound the pattern — keyboard rows, instrument, Slots and Snap all populated — and drew **zero columns**. Five notes sounding, an empty grid, no message, no console error.

| mini | length | columns drawn, before → after |
|---|---:|---|
| `c4@0.2 e4@0.2 g4@0.2 b4@0.2 c5@0.2` | 0.9999999999999998 | **0 → 1** |
| `c4@1.5 e4 g4@0.2` | 2.7 | 2 → **3** |
| `c4@0.5 e4` | 1.5 | 1 → **2** |

The playhead carried the same arithmetic: its clamp was `steps - 1`, which on a 2.7-column pattern is `1.7` — a value no column index can equal — so the highlight never fired on the last column.

**The fix separates two questions `steps` was answering.** `steps` stays the pattern's LENGTH, which is what the writer spells the music from. A new `columnCount` answers how many columns get DRAWN, and joins `headColumn`/`tailColumn` in the one module that decides where a note stops, reading the same private threshold. It is the pattern's whole columns plus any column a note reaches into — so a note sounding in a partial tail column is drawn, and a partial column holding only a rest is not, because nothing can be placed there.

**The guard asks the consequence, not the representation.** "Is `steps` a whole number?" accepts `0.9999999999999998` — it passes every integrality tolerance a reasonable person writes — and the panel goes on drawing nothing. The question has to be "does the count I hand the renderer cover every note?"

**The issue's proposed fix was measured and rejected.** Rounding `steps` up in the reader re-emits `c4@1.5 e4@1.2` as `c4@1.5 e4@1.2000000000000002 ~` — an invented trailing rest that lengthens the pattern. That refutation ships as a test, so it cannot be retried by accident.

## #1089 — the velocity lane offered a drag on writes it always declined

Every column with a note got a `ns-resize` cursor and a pointer handler regardless of whether the gain could be written. On `note("[c5@0.5 f4@0.5 f5@3]")` the bar takes the press and nothing happens — no document change, no movement, no message.

The writer's refusal is **correct**: the gain mini is one slot per column (`"0.5 1 1 1"`), and a note beginning mid-column has no slot of its own. The panel just never asked. Its gate was `gainInScope`, which answers whether the lane should *render*, not whether a drag can *write* — and the two had drifted apart.

**The fix asks the real writer.** `serializeRollGain(model).kind === 'skip'` is already the exact predicate, so this is a read, not a new rule. A `can<Op>` that predicts the writer is how controls end up clickable-but-inert with the whole suite green.

The lane still renders, read-only rather than hidden, and that is deliberate: the bars carry real information — the gains the pattern actually has — they just cannot be changed here. The reason is stated on the columns, where the pointer goes. It needs its own sentence: 30 of these 33 patterns already show *"to add one, use the code view"*, but that is about placement, and a reader takes it to mean notes cannot be ADDED, not that the ones here cannot be made louder.

## Measured

Recorded as committed arms in `cell-coverage.test.ts`, so the figures cannot rot into prose.

- **#1087** — 544 roll models, 0 of fractional length, **0 losing a note**. A real zero, not a dead probe: a control arm in the same run flags 4 hand fixtures and clears the 2 whose weights sum to a whole number. So this notation is reachable by hand and no real material writes it today — which sets the severity, not the validity.
- **#1089** — 424 in-scope models: 391 the writer accepts, **33 it declines**. **305 columns** lose an affordance that never worked; **2950** keep the drag that does. Asking the current model rather than the post-drag one is measured, not assumed — a gain edit does not move a note, so the two answers never differ (0 of 2950).

## Test plan

- editor **3091** / 214 files (baseline 3079/213 — the new arms only)
- app **981** / 73 files, parity-corpus **374** / 28
- tsc **62** editor / **1** app — both pre-existing, unchanged
- browser **68/68** serial (63 baseline + 5 new)
- Each term of `columnCount` was break-tested independently: dropping the note term reddens 5 arms, dropping the length term reddens 3. Neither is defensive duplication.
- Observed in the running app, because both defects are invisible to a unit suite by construction — nothing throws. Screenshots confirm the previously-empty panel now draws all five notes at their true 0.2-column offsets, and the partial-tail note is drawn.

Closes #1087
Closes #1089

Part of #925.


MrityunjayBhardwaj:
Rebased onto `fb1eeb4e` after #1090 landed — `rebase --onto` dropped #1090's two commits and replayed only #1087 and #1089, verified by content (8 files, +616 −42, identical to the pre-rebase diffstat). Base retargeted to `studio_v0.2.0`, so the stacked-parent trap is discharged.

Gates on the rebased tip: editor **3091**/214 · app **981**/73 · tsc **62**/**1** (both pre-existing).

⚠️ **Interacts with #1093, invisibly to git.** The arm here that pins what rounding `steps` would have produced (`'c4@1.5 e4@1.2000000000000002 ~'`) changes its answer once #1093 lands, because #1093 teaches the writer to spell a fractional rest — it becomes `'c4@1.5 e4@1.2 ~@0.3'`. Neither PR touches the other's test file, so both read CLEAN and the second one to merge turns the branch red.

The finding this arm records still holds: the rounded spelling still lengthens the pattern 2.7 → 3 and still moves e4 from `[0.5556, 1.0000)` to `[0.5000, 0.9000)`. Rounding the model field is still wrong; only the shape of the damage changed. Whichever of the two merges second should update the expected string **and** its comment together.

PR #1093, opened 2026-07-29
🎼 The writer spells a rest at its true width, so a drag moves only the note you dragged (#1092)

MrityunjayBhardwaj:
Closes #1092. Part of #925.

Independent of the other four open PRs — nothing here touches `PianoRollGrid.tsx`, `usePlayingStep.ts` or the column-rule work, so this can merge in any order relative to them.

## What was wrong

`@n` is a relative **weight**, so a roll's length and a note's position inside it are not always whole numbers of columns. Every place the writer needed to say *"leave `w` columns of silence here"* said it as a run of bare `~` — and a bare `~` is exactly one column. A gap of 1.5 came back as 2.

That is not a cosmetic rounding. Strudel scales each comma-lane to its own total, so a lane that comes back longer than the pattern puts every lane on a different grid. Dragging `c4` in `note("c4@1.5 e4@1.2")`:

```js
$: note("~ c4@1.5 ~, ~ ~ e4@1.2000000000000002")
```

Lane sums **3.5** and **3.2** where the pattern is **2.7**. Measured through `queryArc`, `e4` — which the user never touched — moved `[0.5556, 1.0000)` → `[0.6250, 1.0000)`. The float is the reader's division re-emitted verbatim into the user's source.

## The issue's scope was too narrow, and that changed the fix

It described this as reachable only on a pattern "whose length is not a whole number of columns". Measured before writing anything:

- **`note("c4@1.5 e4@1.5")` is exactly 3 columns long and was corrupted the same way**, because a note *starts* at 1.5. Its two lanes even came out with **equal sums** (3.5 and 3.5) — so the invariant as filed, *"every lane sums to the pattern length"*, would have read green over it. The rule that actually holds is that every group must land at its **true start**, and the lane must end at the pattern's length.
- A **pure velocity edit** reaches it. No note has to move — `c4@1.5 ~@0.5 e4` with a gain drag re-emitted as `c4@1.5 ~ e4`, and the `~@0.5` the user wrote became a whole column.
- It is not one site. The same whole-column walk lived in the note lanes, in the **gain lane**, and in the region re-emitter. The gain mini runs 1:1 against the note mini, so fixing one and not the other would have put a note's volume on a different column from the note.

## The fix

One question, asked in one place.

- **`columnSplit`** (`model.ts`, beside `columnOverlap`, sharing its private epsilon) — how does a width divide into whole columns and a remainder? "A sliver is not a column" has to mean the same thing to the panel that draws a note and the writer that spells the silence before it.
- **`restTokens`** — `floor(w)` bare rests plus a weighted `~@r`. Whole widths stay bare, so nothing that was already right changes a byte. Its `null` for a *negative* width is load-bearing: `spliceRoll`'s rebuild is the one caller that does not pack its notes first, and its "if its notes now overlap, that isn't a lane" comment is discharged entirely by that refusal.
- **`weightToken`** — rounds at the point of emission only. The model field stays exact, because the panel positions notes from it; it is the document that must not carry `1.2000000000000002`.
- **`reemitRollRegion`** now honours the contract its own docstring already stated — *"a region owning `w` columns must come back as exactly `w / div` steps' worth of weight, **or not at all**"* — by declining a span it cannot tile in whole steps. The caller rebuilds the part as a flat lane, which *can* spell a fractional width, so the edit still lands.

Weighted rests were checked against the engine and against our own reader before being chosen: `~ c4@1.5 ~@0.2, ~ ~@0.5 e4@1.2` reads back to exactly the intended model.

## Measured

| | before | after |
|---|---:|---:|
| fixture gestures written | 102 | **107** |
| …of those, mistimed vs. their model | **44** | **0** |
| fixture gestures declined | 24 | **19** |
| corpus serializations changed (3759 rows) | — | **0** |

44 corrected, 5 gained, 0 lost. The corpus figure is the churn check: every real mini serializes byte-for-byte as it did, both untouched round-trips and a gesture per note.

**Reach on real material is zero, and the gate says so out loud.** 0 of 544 corpus rolls carry a fractional note position and 0 of the 94 flat ones do, so the corpus arm is green before this change and after it — it is a regression floor, not evidence. The evidence is the fixture arm, which fails on every one of the seven fixtures without the fix. `note("c4@1.5 e4@1.2")` is an ordinary thing to write; the severity is low, the validity is not in question.

## Test plan

- `fractionalWidth.test.ts` — 17 arms. **Break-tested per term, not per rule**: each of the seven things this change does was reverted on its own and reddens its own arms (weight rounding → 1, rest remainder → 6, negative-width refusal → 1, trailing pad → 3, gain pads → 1, region guard → 1, epsilon → 1). The first pass showed the negative-width refusal reddening *nothing*; asking what it was for found the `spliceRoll` caller that depends on it, and it now has an arm.
- `lane-width.test.ts` (parity-corpus) — the same comparison over the corpus and the fixtures, **through the engine** rather than by re-reading our own output: every wrong output here re-parses perfectly well. Population, comparison axes and the "why this cannot fire" figure are all pinned.
- `piano-roll-fractional-width.spec.ts` — the real gesture in the real app. Verified to **fail on the base commit** with exactly the reported corruption while its whole-column control stays green.

## Gates

editor **3096**/214 · app **978**/74 · parity-corpus **372**/29 · tsc **62** editor / **1** app (both pre-existing, neither in a changed file) · browser **62/62** serial (60 baseline + 2 new).


MrityunjayBhardwaj:
## Critical self-review

Five things audited on the diff. Two changed the shipped code, three are recorded as clean with the measurement that says so.

### 1. The invariant as filed was not sufficient — this changed the fix

The issue asked for *"every comma-lane sums to the length of the pattern it read"*. Necessary, **not sufficient**, and a gate built on it reads green over the defect: `note("c4@1.5 e4@1.5")` emitted two lanes of **3.5 and 3.5** — equal sums, neither of them the pattern's 3, `e4` moved anyway. The rule that holds is positional: every group lands at its **true column**, and the lane ends at the pattern's length. The sum is a consequence.

This is why `lane-width.test.ts` compares through the engine against plain arithmetic on the model, and never checks the lanes against each other.

### 2. A term that reddened nothing — and was not dead

Break-testing the seven terms one at a time, the negative-width refusal in `restTokens` reddened **zero arms**. Rather than delete it, I asked what it was for: `spliceRoll`'s rebuild is the one caller that does **not** pack its notes first, and its "if its notes now overlap, that isn't a lane" comment is discharged entirely by that `null`. Clamping instead would have emitted overlapping notes side by side as if consecutive. It now has its own arm (stretching a note over its neighbour must reach the comma-lane writer, losing neither note).

Same discipline as #1091's `ceil`-vs-`floor`, opposite outcome: there the term was wrong, here the *question* was missing.

### 3. The rule was written three times, not once

`laneString`, `serializeRollGain`, and `reemitRollRegion` each had their own whole-column walk. The gain mini runs 1:1 against the note mini, so these could not be fixed independently even in principle — weighted rests on one side and bare rests on the other would put a note's volume on a different column from the note. Sixth instance of duplicated semantics at this boundary; all three now ask one helper.

### 4. Checked and clean, with the measurement

- **The step grid does not share the defect, for a principled reason.** Its reader *resolves* fractional weights onto a finer integer grid (`bd@1.5 sd@1.2` → 9 columns, integer cell lengths), so it never has to spell a fractional width; where it cannot resolve, it declines (`bd@1.5 ~@0.5 sd` → null). 0 of 958 corpus grids have a fractional `steps`. The roll differs because it anchors per-note and keeps the fractional start. Boundary drawn right, not an unfixed sibling.
- **`groupWrapRegion` and `rollBars`** were suspected of the same integer walk. Both verified through the engine on fractional fixtures: `[c4@3 e4]@1.5 g4` edits to `c4@3.5 ~ e4@1.5 g4@4` (sums to its 10 columns, every onset correct), and the multi-bar rebuilds either round-trip byte-identical or decline. No note dropped.
- **The leaf path is untouched** — it does byte surgery and emits no widths.

### 5. One cosmetic residual, not fixed, stated rather than left implicit

A remainder in `(1e-9, 1e-6)` would stringify exponentially — `~@1e-7` in the user's source. krill parses that **exactly** (`c4@1e1` is identical to `c4@10`, verified), so it is correct, merely unusual-looking, and the size is below anything musical: the floor is `COLUMN_EPS` and real weights are orders of magnitude larger. Not worth code; recorded so it is a known shape rather than a surprise.

### Merge note

`model.ts` will conflict textually with #1090/#1091, which add `headColumn`/`tailColumn`/`columnCount` in the same region — `columnSplit` is deliberately their sibling, beside `columnOverlap` and sharing the same private epsilon. Small, and semantic conflict is nil; nothing else in this PR touches a file those branches change.

MrityunjayBhardwaj:
⚠️ **Ordering note — this PR and #1091 interact, and git can't see it.**

Rebased onto `fb1eeb4e` (post-#1090). Gates on the rebased tip: editor **3096**/214 · app **982**/74 · tsc **62**/**1** (both pre-existing).

The thing worth flagging: #1091 and this PR are each CLEAN on their own, but landing both leaves `studio_v0.2.0` red. #1091's `columnCount.test.ts` pins the counterfactual — what rounding `steps` in the reader *would* have produced — as `'c4@1.5 e4@1.2000000000000002 ~'`. This PR teaches the writer to spell a fractional rest, so that same counterfactual now comes out `'c4@1.5 e4@1.2 ~@0.3'`. The two branches share no test file, so nothing conflicts and nothing warns.

**#1091's conclusion is unaffected** — I checked whether the rounded spelling still corrupts, and it does: the pattern still goes 2.7 → 3 columns and e4 still moves from `[0.5556, 1.0000)` to `[0.5000, 0.9000)`. Only the *mechanism* changed, from a distorted weight to a correctly spelled rest. So the pin wants re-arguing, not just re-numbering.

Suggested order: **merge #1091 first**, then I'll rebase this one and update that pin with its reasoning in the same commit. Merging this first works too — #1091 then needs the same edit.

